PR #10, opened 2020-02-26
Port iomux to imxrt-ral

bfrog:
Note that each pad requires unsafe modify_reg() calls to convert to a
particular pads alternate usage.

This should be safe because each pad has its own register pair (mux/pad) for
setting things up. So long as the only place where these registers are
used is in the iomuxc instance in struct Peripherals, this is like just
fine.

I'm not excited to have to import the modify_reg!() macro in each gpio module, I didn't try using it using crate::ral::modify_reg!() which perhaps works already, will try later perhaps.

### DIFF
--- a/imxrt-hal/Cargo.toml
+++ b/imxrt-hal/Cargo.toml
@@ -26,15 +26,17 @@ test = false
 
 [features]
 default = ["embedded-hal/unproven"] # Allows us to access the new digital pin traits
+
+# Device specific feature flags
 # these need fixes and conditional sections in CCM
 #imxrt1011 = ["imxrt-ral/imxrt1011"]
 #imxrt1015 = ["imxrt-ral/imxrt1015"]
 #imxrt1021 = ["imxrt-ral/imxrt1021"]
-imxrt1051 = ["imxrt-ral/imxrt1051"]
-imxrt1052 = ["imxrt-ral/imxrt1052"]
-imxrt1061 = ["imxrt-ral/imxrt1061"]
+#imxrt1051 = ["imxrt-ral/imxrt1051"]
+#imxrt1052 = ["imxrt-ral/imxrt1052"]
+#imxrt1061 = ["imxrt-ral/imxrt1061"]
 imxrt1062 = ["imxrt-ral/imxrt1062"]
-imxrt1064 = ["imxrt-ral/imxrt1064"]
+#imxrt1064 = ["imxrt-ral/imxrt1064"]
 rtfm = ["imxrt-ral/rtfm"]
 rt = ["imxrt-ral/rt"]
 nosync = ["imxrt-ral/nosync"]
--- a/imxrt-hal/src/iomuxc.rs
+++ b/imxrt-hal/src/iomuxc.rs
@@ -37,6 +37,9 @@ pub struct Alt8;
 pub struct Alt9;
 
 pub struct IOMUXC {
+    // private keep this instance so it is taken
+    iomuxc: crate::ral::iomuxc::Instance,
+
     //
     // GPIO_B0
     //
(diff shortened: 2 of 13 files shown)


imxrt-hal/src/iomuxc/macros.rs
```diff
@@ -2,13 +2,9 @@ macro_rules! alt0 {
     ($Pad:ident, $mux_mod:ident) => {
         /// Converts the pad into its Alt0 setting
         pub fn alt0(self) -> $Pad<$crate::iomuxc::Alt0> {
-            self.iomuxc().$mux_mod.modify(|_, w| {
-                w.mux_mode()
-                    .variant($crate::ral::iomuxc::$mux_mod::MUX_MODE_A::ALT0)
-            });
+            unsafe { modify_reg!($crate::ral::iomuxc, $crate::ral::iomuxc::IOMUXC, $mux_mod, MUX_MODE: $crate::ral::iomuxc::$mux_mod::MUX_MODE::RW::ALT0) };
```
mciantyre: The `use imxrt_ral::modify_reg;` imports throughout the `gpio` files could be removed if the `modify_reg!` macro were absolutely referenced. How about

```
::imxrt_ral::modify_reg!($crate::ral::iomuxc, ...)
```

instead of `modify_reg!`? Comment applies for each `modify_reg!` usage in this file.

imxrt-hal/Cargo.toml
```diff
@@ -26,15 +26,23 @@ test = false
 
 [features]
 default = ["embedded-hal/unproven"] # Allows us to access the new digital pin traits
+
+# Device level features to allow for simpler conditional compiles
+imxrt101x = []
+imxrt102x = []
+imxrt105x = []
+imxrt106x = []
```
mciantyre: Will these be something others will depend on, and we'll need to support going forward? Or, are these just for our internal use?

If it's the latter, we should document that these are not for crate users. Maybe we prefix them with a `_`?

We could try something like the tokio team, where they have [internal configuration macros](https://github.com/tokio-rs/tokio/blob/master/tokio/src/macros/cfg.rs) to express combinations of features. The macros are [used to selectively enable modules](https://github.com/tokio-rs/tokio/blob/master/tokio/src/lib.rs#L303), or functions, or structs. With this approach, all feature-flags are usable by users, so there's no ambiguity. And, we can replace things like

```rust
#[cfg(any(feature = "imxrt1060", feature = "imxrt1062", feature = "imxrt1064"))]
pub mod foobar { /* ... */ }
```

with

```rust
cfg_imxrt106x! {
    pub mod foobar { /* ... */ }
}
```

mciantyre:
See my thoughts on the need to import `modify_reg!` throughout the `gpio` modules.

This all LGTM! Built with `cargo build --features imxrt1062`.

Mind running `cargo fmt`?

imxrt-hal/src/iomuxc.rs
```diff
@@ -81,59 +85,64 @@ pub struct IOMUXC {
...
+
     // GPRs
-    //
     pub gpr: GPR,
 }
 
 impl IOMUXC {
-    pub(crate) fn new(iomuxc: crate::ral::iomuxc::IOMUXC) -> Self {
+
+    #[cfg(feature = "imxrt106x")]
```
mciantyre: Should this `#[cfg(feature = "imxrt106x")]` go on top of the `impl IOMUXC`? It looks like we have another one on top of the `struct IOMUXC` declaration. Would the possibly-empty `impl IOMUXC` compile without a `struct IOMUXC`?

bfrog: Yeah I wasn't sure how I wanted to deal with this exactly, I want to maybe even import a different iomux per family, like in lib.rs 
```
#[cfg(feature = "imxrt106x")]
use iomuxc/imxrt106x as iomuxc;

So then the peripherals struct doesn't have to deal with it, similar to how ral exposes things based on feature flags. It's something I thought was worth discussing. Let me know what you think on that.

imxrt-hal/Cargo.toml
bfrog: The macro wrapper is a neat idea, I'll take a closer look at that. That's the general idea though, to internally allow maximum code sharing without typing massive #[cfg()] options

imxrt-hal/src/iomuxc/macros.rs
bfrog: Perfect, yeah I'm still so used to pre-2018 where everything had to be extern use to get macros in, I never fully learned the new rules beyond use xyz::somemacro;

Neat, will fix

bfrog:
I'm going to skip the features for conditional compilation on this for now since the goal is to support the 1062 on the teensy and the 1060evk to start. I think we can revisit it in a nicer way perhaps once I do some work to support the 1010evk (imxrt1011). I don't have any other evks at the moment to test against so perhaps we can leave that up to others as they wish to contribute that work, with the way paved by the 3 boards and 2 chips being supported.

bfrog:
Thanks for the review @mciantyre

PR #11, opened 2020-02-29
Port I2C

mciantyre:
The PR updates the I2C peripheral to use the RAL API. The port was fairly mechanical: replace PAC functions with the corresponding RAL macros. We should review it for consistency.

There are some other changes that are bundled in this PR. These changes might have an effect on how we port other peripherals.

- Remove the `SW_PAD_CTL_PAD_*` modules from all the `pad!(...)` macros. It doesn't seem that we can use the RAL macros with RAL types; the RAL macros only work on symbols. I'm not sure there's value in provided a `pad()` method on a pad type. I propose that we call `write_reg!` in a pin's `configure()` method. The `iomuxc/i2c.rs` changes makes this proposal concrete.
- Remove the `IntoDaisy` trait and the `daisy` module. It's not necessary; we may configure input selection in the pin's `configure()` method. The `iomuxc/i2c.rs` changes demonstrate this approach.
- Remove some of the strong types around timing, notably the `Ticks` type. It's easier to work with the underlying primitive types, and the `Ticks` type should have stayed as my experiment in the previous HAL implementation 😝

I've not tested on hardware. I may try to integrate the HAL into the Teensy 4 BSP so that I can test it. But, I'm pretty confident that we can catch any issues by inspection.


imxrt-hal/src/iomuxc/i2c.rs
```diff
@@ -33,193 +35,185 @@ pub trait Pin {
     type Module: module::Module;
 
     /// Perform IOMUXC configurations for this pin
-    fn configure(&mut self);
+    fn configure(self);
```
bfrog: Going from &mut self to self seems like something I should look harder at, if we consume the pad generated by the pad macro, does this mean it is forever an i2c iomuxed pad then?

I guess I'm thinking about low power devices where I want to perhaps dynamically setup and tear down pins/busses/devices. Something to perhaps look harder at in the future, not that concerned while we ramp up here.

mciantyre: Good thought about peripheral reconfiguration. Right now, the HAL API doesn't accommodate for any reconfiguration. This pad is forever muxed for I2C purposes. We further enforce this by taking the pads by value in the `i2c::Builder::build()` method. The returned `I2C` object doesn't provide a `release()` method to return the pads. We can figure out how this might change, not only for the pads and peripherals, but also for the clocks that drive the peripherals.

Issue #8, opened 2020-02-26
Port HAL to the RAL


mciantyre:
The issue tracks how much of the HAL is migrated from the PAC-based implementation to the RAL-based implementation.

- [x] CCM
- [x] GPIO
- [x] I2C
- [x] IOMUXC
- [x] PIT
- [x] PWM
- [x] SPI
- [x] UART

My focus is on keeping parity with the existing features; for example, I don't intend on generalizing the HAL for other iMXRTs right now. Once I can build the HAL, I can show equivalence by [using the examples in the `teensy4-examples` crate](https://github.com/mciantyre/teensy4-rs/tree/master/teensy4-examples/src).

mciantyre:
IOMUXC is WIP on `hal_iomux`

mciantyre:
Ported IOMUXC in #10

bfrog:
Initial iomuxc work done and merged, on to the next thing. I plan on trying to deal with GPIO stuff this week so I can get the typical blinky led example working on the 1060evk. I also picked up a teensy so I can start testing the teensy examples you've made.

mciantyre:
Awesome thanks. I can start the I2C peripheral unless there’s something else you’d like to see.

bfrog:
I2C sounds great, will be super happy when we get a uart going as well :+1:

mciantyre:
I2C port proposed in #11.

bfrog:
GPIO in progress, but noticed I think a case where our unsafe usage really is in fact unsafe, see #21. Something I think we'll need to consider what each option means

bfrog:
GPIO wrapped up, though future improvements to iomux+gpio are definitely possible to better take advantage of rusts ownership and mutability rules to avoid the critical section

bfrog:
@mciantyre did you want to try to tackle pwm or should I? I can hammer it out here I think pretty quick, don't want to step on your toes if you are already on it

mciantyre:
I haven't started on PWM yet, but I'm happy to pick it up. I might not get to it until this weekend.

PWM might be the least straightforward to port. When we implemented the driver on the `svd2rust` output, we were taking advantage of SVD `<cluster>`s being represented as a struct. But, the RAL approach doesn't represent SVD clusters (#2); it simply duplicates the registers. This might require us to change a good bit of the implementation; it might be more macro-heavy. There might be some parts of the RAL API that I'm not considering, so I'm going to study up on that.

[...]

PR #12, opened 2020-03-01
Port UART

mciantyre:
Another peripheral port from the PAC implementation to the RAL implementation. This one focuses on the UART peripheral. It was another mechanical port: change PAC calls to RAL calls. We should review it for consistency. I've not tested it on hardware.

### DIFF
--- a/imxrt-hal/src/uart.rs
+++ b/imxrt-hal/src/uart.rs
@@ -6,7 +6,7 @@
 
 use crate::ccm;
 pub use crate::iomuxc::uart::{self, module};
-use crate::pac;
+use crate::ral;
 use core::marker::PhantomData;
 
 /// An uninitialized UART peripheral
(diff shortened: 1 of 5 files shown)


mciantyre:
Part of #8.

imxrt-hal/src/uart.rs
```diff
@@ -76,61 +76,87 @@ impl Unclocked {
...
-                .uart_clk_podf()
-                .variant(prescalar)
-        });
+        ral::modify_reg!(
+            ral::ccm,
+            ccm,
+            CSCDR1,
+            UART_CLK_SEL: (clock_select as u32),
+            UART_CLK_PODF: (prescalar as u32)
+        );
```
mciantyre: Noting the change from a `write` to a `modify`. Previous implementation was incorrect, and it would have squashed the values in other fields of `CSCDR1`.

PR #14, opened 2020-03-01
Port PIT timers

mciantyre:
The PIT timers needed some internal changes to support the lack of register clustering in the RAL. We extend the existing `Channel` trait to provide implementations of the register manipulations that we need. We then call onto those trait methods in the `PIT` implementation. We should check the port for consistency.


mciantyre:
Part of #8.

PR #22, opened 2020-03-04
GPIO ported

bfrog:
This is a minimal port of what was in the gpio module, but needs to be made safe.


bfrog:
So I looked closer at how the stm Hal is written here and I think a few changes would probably be beneficial. The gpio port handling is especially interesting as they do avoid the issue by using safe rust rules to avoid the issue. For imxrt the separation of the iomuxc and pads means perhaps some more rethinking is needed on that front to make it work correctly with awkward issues. For now I'm going to wrap this in a critical section from cortex-m as it's safe and correct then. I'd rather release a version that matches the API of your teensy Hal and then iterate on improvements, even if drastic, because we might have a few samples to see how the API looks and works in practice better.

Issue #21, opened 2020-03-04
CAS loop or disable interrupts on unsafe modify of shared registers


bfrog:
A few places I'm finding we are doing unsafe modify_reg(); These need to be reviewed in the hal to ensure that when doing an unsafe modify that the register isn't possibly modified elsewhere in the hal.

This is so that after a read but before a write for a modify operation a pre-empted ISR which also modifies the same register may cause a data race once returned to the previously interrupted code.

In such cases (GPIO !) we need to likely either temporarily disable interrupts before and re-enable after these operations *or* implement a compare and swap loop such that when interrupts to occur an atomic cas will fail and a loop may retry the whole task.

This seems like its more likely an issue in the GPIO code where the GPIO port registers are shared among multiple individual GPIO pins, unless we change the way the GPIO is used perhaps to be on a port by port basis with a gpio port singleton.

bfrog:
Fixed for now using critical sections (cortex_m::interrupt::free) but we want to use ownership and mutability rules in the future, so closing this and opening a new issue related to that

Issue #9, opened 2020-02-26
Port SPI implementation from Teensy 4 project to imxrt-hal


mciantyre:
mciantyre/teensy4-rs#47 provides a SPI master implementation. Once it's finished, port the peripheral and the other code changes into the `imxrt-hal`.

mciantyre:
Finished with #24

PR #24, opened 2020-03-15
Port PWM peripheral, add SPI peripheral, support Teensy 4 integration

mciantyre:
The PR ports the PWM peripheral to the RAL, adds the SPI peripheral to the project (#9), and updates the RAL / HAL to build with the `teensy4-rs` project. With this PR, and the [`feature/imxrt-hal` branch of the `teensy4-rs` project](https://github.com/mciantyre/teensy4-rs/tree/feature/imxrt-hal), we can build and execute all examples in `teensy4-rs` on a Teensy 4. It's the culmination of all #8 efforts.

There's a couple of commits to call out. The commit messages may provide additional context.

- `b4301ab` removes the `cortex-m-rt` dependency from the auto-generated RAL. We can't use the general `cortex-m-rt` crate. Right now, users will need to link in their own runtime crate that can support iMXRT memory layout and startup requirements. I'm still using the `imxrt1062-rt` crate in the `teensy4-rs` project to meet this requirement. The [`imxrt-rt` crate](https://github.com/imxrt-rs/imxrt-rt) may also work.
- `0cff649` extends the `imxrtral.py` script to propagate register sizes into fields. See that commit for more information. If accepted, we should consider porting this commit to the `stm32ral` project.
- `2c06800` ports the PWM peripheral. #21 found that some behaviors of the HAL were unsafe. As I ported the PWM peripheral, I found that the PWM peripheral was equally unsafe. The new API should enforce safety while maintaining an implementation of `embedded_hal::Pwm`. Note that the previous HAL exposed an implementation of `embedded_hal::PwmPin`, so this represents a breaking change. The `teensy4-rs` project already accounts for this. I'm looking for some feedback on this API; I'm concerned it might be harder to use. However, it seemed to be one of the only ways to support an `embedded_hal` implementation without baking-in all of the pre-emption safety into the implementation.

If you have a Teensy 4 available, I recommend pulling-down this PR, pulling-down the [`feature/imxrt-hal` branch of the `teensy4-rs` project](https://github.com/mciantyre/teensy4-rs/tree/feature/imxrt-hal), and running the examples on hardware. I still need to measure some timings, but at a first glance all behaviors seem consistent.

Also, @dflemstr deserves the credit for the SPI implementation (see mciantyre/teensy4-rs#47). I've only ported it to use the RAL.


mciantyre:
Note that, since the RAL is no longer in VCS (#16), you'll need to re-generate the RAL to realize the changes in this PR.

PR #34, opened 2020-03-28
CI build using GitHub Actions

bfrog:
Setting up a github actions based CI workflow for this repo seems pretty important at this point to me as I've already failed a few times to ensure fmt, clippy, etc are all happy.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,31 @@
+name: All Checks
+
+on: [push, pull_request]
+
+jobs:
+  build-ral:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        feature: ["imxrt1011", "imxrt1015", "imxrt1021", "imxrt1051", "imxrt1052", "imxrt1061", "imxrt1062", "imxrt1064"]
+    steps:
+    - uses: actions/checkout@v2
+    - name: install virtualenv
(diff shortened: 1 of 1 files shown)


mciantyre:
If it helps, I've set up similar workflows in the [`imxrt-boot-gen` repo](https://github.com/imxrt-rs/imxrt-boot-gen/blob/master/.github/workflows/rust.yml). It [notes `clippy` warnings inline](https://github.com/imxrt-rs/imxrt-boot-gen/pull/2/files), and it [requires `rustfmt`-formatted code](https://github.com/imxrt-rs/imxrt-boot-gen/runs/477224072).

`clippy` reports many warnings when it checks the RAL. We may need a crate-level `#![allow(clippy::all)]` in the RAL to prevent the noise.

bfrog:
I in fact based this on the actions for imxrt-boot-gen, but its a little quirky because of the workspace. I'll have to look maybe at tokio or other workspace projects to get some ideas.

bfrog:
So unfortunately uses doesn't allow a working directory, and the actions-rs/clippy action would need a working directory to work.

There's also a multitude of doc test errors in ral that probably should be fixed, I'll note both as issues and move on. I think having something here that builds, runs tests, and checks formatting is a good start. Adding clippy would be a welcome addition, will note that as another issue.

PR #37, opened 2020-03-29
Enable clippy in CI

### LINKED ISSUE #35: CI Clippy Integration
Currently github actions uses doesn't support a working directory, and the actions-rs/clippy action would need a working directory to work correctly in this workspace repository. I'm not too keen on splitting the repo as its already a chore to modify Cargo.toml's to use local paths for the bsp and rt crates.

Perhaps we can at least run clippy using a github action run task in place of the nicer uses task.

mciantyre:
The PR enables `clippy` in CI. It asserts that there are no `clippy` warnings or errors in the HAL. The PR also cleans up existing `clippy` warnings.

#34 added automated builds and code formatting checks. It did not include a check for `clippy` issues. #35 notes that we can't immediately use the `actions-rs/clippy` GitHub action, since we don't have a working directory for the RAL. I propose that we run `clippy` even if we cannot realize the output in a GitHub UI. The output provided by the job is still actionable.

I structured the commits to show that [the CI flow will catch clippy issues in the HAL](https://github.com/mciantyre/imxrt-rs/runs/543625866).

Closes #35.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -59,3 +59,5 @@ jobs:
       run: cd imxrt-hal && cargo test --verbose --features ${{ matrix.feature }}
     - name: Check format (${{ matrix.feature }}) for HAL
       run: cd imxrt-hal && cargo fmt --all -- --check
+    - name: Run clippy (${{ matrix.feature }}) for HAL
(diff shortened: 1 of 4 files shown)


mciantyre:
Noting that I changed the `imxrtral.py` RAL generation script. The change prevents `clippy` from evaluating the RAL. If you run `clippy` locally and notice issues in your RAL, re-generate the RAL.

PR #40, opened 2020-04-03
Set default clock low power mode to run

bfrog:
SysTick *does not* wake up the chip by default when its put to sleep.

Other useful timers (PIT) probably do, but SysTick does not. This was
highly confusing to discover and left me staring at my debugger screen
for far longer than I'd like to admit.

### DIFF
--- a/imxrt-hal/src/ccm.rs
+++ b/imxrt-hal/src/ccm.rs
@@ -6,6 +6,8 @@ use arm_clock::set_arm_clock;
 use core::time::Duration;
 use imxrt_ral as ral;
 
+use ral::modify_reg;
+
 pub struct Handle {
     pub(crate) base: ral::ccm::Instance,
     pub(crate) analog: ral::ccm_analog::Instance,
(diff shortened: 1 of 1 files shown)


imxrt-hal/src/ccm.rs
```diff
@@ -65,13 +83,24 @@ impl PLL1 {
...
-        CCM {
+        let mut ccm = CCM {
             handle: Handle { base, analog },
             perclk: perclk::Multiplexer::new(),
             pll1: PLL1::new(),
             pll2: pll2::PFD::new(),
             pll3: pll3::PFD::new(),
-        }
+        };
+        ccm.set_mode(ClockMode::Run);
```
mciantyre, line 93: Should changing the low-power clock behaviors be the responsibility of the HAL? This change imposes our beliefs on what the processor should do, rather than representing the behaviors of the hardware. Would it be troublesome to enable this in a BSP, or let end-users enable it?

I rely on SYSTICK not breaking `wfi` in the Teensy 4. This change may also affect `teensy4-rs` users who expect similar behaviors. If we keep this change, I'll need to undo it in the Teensy 4 BSP.

I recommend that we document which default processor behaviors we're changing in the HAL. We should feature these changes in the crate-level documentation. Potential HAL users who are already familiar with the default behaviors of iMXRT chips should know how this library differs from what's available in the reference manual.

bfrog:
Done on master

Issue #44, opened 2020-04-09
Cannot build 0.2.0 crates from crates.io


mciantyre:
I'm hitting build errors when referencing the 0.2.0 releases from crates.io. It affects the RAL, and maybe the HAL.

Minimal reproduction of the RAL:

```
cd /tmp
cargo new --lib test-ral
cd test-ral
echo 'imxrt-ral = { "version" = "=0.2.0", features = ["imxrt1062"] }' >> Cargo.toml
cargo build

...

error: couldn't read /Users/mciantyre/.cargo/registry/src/github.com-1ecc6299db9ec823/imxrt-ral-0.2.0/src: Is a directory (os error 21)
```

It looks like 742e687 explicitly sets the RAL library's path to `src`. Does that need to be the path to the library entrypoint, `src/lib.rs`? Is there a reason we're explicitly setting the library path?

Additionally, the [HAL references a local directory containing a 0.2.0 release of the RAL](https://github.com/imxrt-rs/imxrt-rs/blob/742e68757abc9d1a0595ab3105c24e9aa8f27497/imxrt-hal/Cargo.toml#L17). Should the path be there? I'm actually not sure if that will be troublesome for end users.

mciantyre:
I'm playing around with a `0.2.1-alpha` release of both crates. Goal is to download the HAL from crates.io in the `teensy4-rs` project.

mciantyre:
Wanted to see if this was system-dependent, and it seems that it is! I cannot reproduce this in a Linux VM.

mciantyre:
We may want to consider yanking the 0.2.0 release. As of this writing, I've not yanked it. I'd like for someone else to reproduce the issue before we consider a yank.

mciantyre:
I've release 0.2.1 of both the RAL and HAL to crates.io. See #45.

bfrog:
Thanks! I added that src because cargo publish was complaining about not being able to find src/lib.rs, not sure why. I fixed that, only to run into other issues, said forget it and can publish --no-verify, thanks for fixing this!

bfrog:
We should probably yank 0.2.0

PR #45, opened 2020-04-09
Release 0.2.1

mciantyre:
See #44. I can't build the `0.2.0` release from crates.io. The PR represents the state of the repo when I published a `0.2.1` release.

I started by publishing a `0.2.1-alpha` release. The alpha release was needed to put us in a good state to build and release `0.2.1`. While stepping through the `0.2.1` release, I wrote some notes on how I did the release. I put the notes at `docs/RELEASE.md`.

Show that both crates build when referenced from crates.io:

RAL:
```
cd /tmp
cargo new --lib test-ral
cd test-ral
echo 'imxrt-ral = { "version" = "=0.2.1", features = ["imxrt1062"] }' >> Cargo.toml
cargo build
``` 

HAL:
```
cd /tmp
cargo new --lib test-hal
cd test-hal
echo 'imxrt-hal = { "version" = "=0.2.1", features = ["imxrt1062"] }' >> Cargo.toml
cargo build
```


mciantyre:
0.2.1 release works in the `teensy4-rs` project. I didn't supply a tag. But, the actual release occurred at 28c995a.

mciantyre:
A change to a markdown file (3a55171d5fe0b55585be67c165f2940bfa8be211) broke the build... 😮

mciantyre:
Nevermind, I tried harder, re-ran the jobs, and they all pass now 😛I'll keep an eye on that.

mciantyre:
I might have introduced CI flakiness by removing the RAL's Cargo.toml file. Here's [the first batch of jobs that failed](https://github.com/mciantyre/imxrt-rs/runs/572554517) on 3a55171. Does the cache not include the auto-generated Cargo.toml file? Let me know what you think; I'm happy to revert a95bf24 if we need.

imxrt-hal/Cargo.toml
```diff
@@ -9,12 +9,12 @@ keywords = ["imxrt", "nxp", "embedded", "no_std"]
 categories = ["embedded", "no-std"]
 license = "MIT/Apache-2.0"
 edition = "2018"
-version = "0.2.0"
+version = "0.2.1"
 
 
 [dependencies]
 cortex-m = { version = "0.6" }
-imxrt-ral = { version = "0.2.0", path = "../imxrt-ral" }
```
bfrog, line 17: We shouldn't need to remove the path, but we do need to change the version

bfrog: See https://github.com/tokio-rs/tokio/blob/master/tokio-tls/Cargo.toml for example

PR #49, opened 2020-04-23
Add DMA channels

### LINKED ISSUE #43: Support SPI DMA transfers
From @dflemstr:

> I love the recent development and buzz around SPI support. In my application, I want to transfer a lot of data, and it would hence be useful to support DMA transfers for SPI!
>
> I have another side project where I've been trying to eek out the maximum amount of performance from SPI in a device independent manner. I defined the API to look like this: https://github.com/dflemstr/embedded-platform/blob/master/src/spi.rs
> 
> Of course, that introduces async as well, while this crate/embedded_hal is very focused on blocking APIs. However the main take-away from that API I think is:
> 
> You initiate a transfer which takes ownership of the slice of memory you want to send
You can poll that transfer to completion (and do that in a tight loop if you want a "blocking" API)

Manually migrating this from mciantyre/teensy4-rs#51

mciantyre:
The PR adds DMA channels, and a set of traits that lets us add DMA capabilities for iMXRT peripherals. I extend the UART peripheral with the new DMA capabilities.

There's [a rough example in the `teensy4-rs` repo](https://github.com/mciantyre/teensy4-rs/blob/feature/dma/teensy4-examples/src/dma_uart.rs) showing how we can use DMA to coordinate UART transfers and receives. It needs to be cleaned up and documented, but it seems to work!

There are a bunch of DMA capabilities that this PR does not support, like

- channel arbitration modes. The DMA controller supports fixed-priority (default) or round-robin transfer scheduling. Users can't change the default right now.
- group arbitration. There are two DMA channel groups, and we could modify how the DMA controller schedules transfers across / within the two groups. Users can't do anything with groups right now.
- changing channel priorities. This might be the simplest to implement: we should let user's swap the priorities of two channels. This would let us keep the requirement that no two channels have the same priorities. Priorities also includes channel preemption. Users can't change channel priorities or preemption settings.

If you think we need these DMA capabilities, or have ideas on how we can add them in, let me know!

The PR closes #43 once we extend the SPI peripheral with the new DMA capabilities, and we test SPI DMA transfers.


mciantyre:
Added SPI DMA transfer and receive capabilities. [The `dma_spi.rs` example in the `teensy4-rs` repo](https://github.com/mciantyre/teensy4-rs/blob/cf6d85cc1aac4e60e37e3e316a359bcda1b8b9fd/teensy4-examples/src/dma_spi.rs) is a contrived example of reading one byte from a SPI slave using DMA.

bfrog:
This is super, I need to look at the DMA controller on this chip closer before I can really comment well on this, will do my best to get a thorough review here soon-ish

mciantyre:
There's [an enhanced UART DMA example](https://github.com/mciantyre/teensy4-rs/blob/0ba381b8c30e04005bb3ae7e50b579c199f97679/teensy4-examples/src/dma_uart.rs) in the `teensy4-rs` test suite. I also updated the [SPI DMA example](https://github.com/mciantyre/teensy4-rs/blob/0ba381b8c30e04005bb3ae7e50b579c199f97679/teensy4-examples/src/dma_spi.rs) to perform full-duplex, multi-`u16` transfers with a connected IMU.

mciantyre:
I think this is ready for review. There's two working examples in the `teensy4-rs` repo (see links above). I added a longer doc example in the `dma` module, which should be checked by our tests.

SPI and UART are DMA-capable peripherals. If we want any other DMA-capable peripherals, or any other DMA features, let me know! I might play with I2C and DMA, but unless someone has a real need for DMA-capable I2C, I don't consider it a blocker for DMA.

bfrog:
Fantastic, I'll spend sometime this afternoon reviewing. This is fantastic, and will certainly be needed once I get to the SAI and USB audio stuff, which I want to use on a board design I've been working on

imxrt-hal/src/dma/element.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+impl Element for u32 {
+    const DATA_TRANSFER_ID: u16 = 2;
+}
+
+impl Element for u64 {
+    const DATA_TRANSFER_ID: u16 = 3;
+}
+
```
bfrog, line 28: I suppose to support the 32-byte burst feature here something perhaps a little more special might be needed? Would likely need to be setup to look like 

```impl Element for u8[32]```
or
```impl Element for u16[16]```

And so on down the line of uint types I'd guess?

bfrog:
First once again, I want to say how awesome of an addition this is. That said I see a few things that could perhaps be improved through either inline commentary *or* self describing function names that perhaps better describe whats going on.

Particularly the 4 functions that setup the TCD registers with source/destination addresses, offsets, last address "add" values, things like that. The ref manual doesn't really describe why its setting all the registers itself in its examples, I had to look at the register descriptions. That's a lot of in depth digging to better understand the ~5-10 lines of register manipulation that happen there. It'd be nice if things were self descriptive in the code there.

I think overall this is a great first round on DMA. There's features like you mentioned that are missing. One feature I particularly like in this DMA controller is its ability to automatically read or write from a ring buffer using the MOD field. Its not quite as awesome as "here's my N buffers, read from them or write to them" as the buffer must be contiguous in memory, but still something I'd likely use myself in the future. It's not entirely clear how things might change to use that feature here. the set_source/set_destination don't likely need it, but perhaps the set_destination_buffer, set_source_buffer could.

Lastly a DMA powered memcpy is always a nice addition, and this DMA controller seems to support it.

It'd be nice to have a dma::memcpy(src, dest) with one day perhaps a Future return type, a pipe dream perhaps.

mciantyre:
Thanks for the review. The latest update

- adds developer docs to describe the `set_source()`, `set_destination_buffer()`, and related methods.
- adds a `buffer` module, which lets users allocated and share DMA-safe source and destination buffers. It supports normal arrays, which we're calling `Linear`. It also supports circular buffers, which we're calling `Circular`. The types in the `buffer` module let us remove the `unsafe` methods from the `Peripheral` interface.
- adds a `Memcpy` type (untested), which should transfer memory across two buffers using DMA
- updates the docs.
- cleanups and method renames.

The `development` branch in the `teensy4-rs` repo has up-to-date examples of the new DMA API. `Memcpy` example WIP.

imxrt-hal/src/dma/element.rs
mciantyre: The `SSIZE` documentation in the reference manual says that the burst is "4 beats of 64 bits." I'd be partial to _only_ support `[u64; 4]`, as it describes the "beats" of the burst.

Am I misinterpreting the 32-byte burst?

mciantyre:
Update to the update:

- refactor the DMA channel methods, leading to an approach that's easier to document, and can support both peripherals and memcpy transfers
- test DMA-powered memory copy. The first, untested attempt didn't work. It seems to work now, and there's [an example](https://github.com/mciantyre/teensy4-rs/blob/b35a445dcdec8195297544624625bb1036848ff8/teensy4-examples/src/dma_memcpy.rs) in the Teensy 4 repo, which transfers data across two circular buffers. See the commits for justification.
- fix run-time checks for circular buffer alignment; fix modulo computation for circular buffers

I think this address all thoughts from the first review. Ready for round two!

bfrog:
I'm fine with this, a few nitpicks don't block a merge in my mind. Fine to merge when you are ready

mciantyre:
Sounds good. I'll do a self-review, and merge this branch within a week. Let me know if there's anything else that might help with SAI support.

PR #51, opened 2020-04-25
Add development instructions to Readme

Walther:
Suggestion: add development instructions in the readme, to make contributing easier.
Let me know if there's more steps required!

Before running these steps, I wasn't able to run `cargo check` in the workspace:
```
cargo check
error: failed to read `/Users/walther/git/imxrt-rs/imxrt-ral/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```

After these steps, I can run `cargo check`, but I get 8324 errors 😅 

[errors.txt](https://github.com/imxrt-rs/imxrt-rs/files/4533768/errors.txt)

```
error: aborting due to 8324 previous errors

Some errors have detailed explanations: E0425, E0432, E0433.
For more information about an error, try `rustc --explain E0425`.
error: could not compile `imxrt-hal`.
```

PR #52, opened 2020-04-25
Fix CI builds

mciantyre:
The PR addresses issues with our CI builds. We excluded the RAL's `Cargo.toml` in #45, and we seem to regularly see CI build failures after that PR. The errors indicate that the system [can't find the RAL's `Cargo.toml`](https://github.com/imxrt-rs/imxrt-rs/runs/618287558), which we were expecting to be in VCS (example error in #51).

We now cache the entire `imxrt-ral` directory, not just the `imxrt-ral/src` Rust source files. This includes the RAL's `Cargo.toml` in the cache.

I also swapped some build steps. If we have a cache hit, we don't need to install the Python dependencies and set up the Python environments, since they're only necessary to build the RAL. It lets us skip two more steps for each cache hit.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,7 +21,7 @@ jobs:
       uses: actions/cache@v1
       id: ral-cache
       with:
-        path: imxrt-ral/src
+        path: imxrt-ral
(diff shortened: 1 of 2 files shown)


mciantyre:
Example of skipped Python installations, and RAL generation, after warming the cache. We should expect this to be our steady-state.

Screen Shot 2020-04-25 at 5 59 08 PM

PR #54, opened 2020-04-29
Update README with usage and Q/A; add CONTRIBUTING docs

### LINKED ISSUE #53: Add usage and contribution documentation
We've found that it's not easy for new contributors to get started (#51). We should add some getting started documentation, and some contribution documentation.

Things we should cover

- how to use `imxrt-hal` in another project
  - note that we **require** a feature flag
- how to contribute to the project
  - how to generate the RAL
  - helpful external resources (reference manuals, other projects to study)
- how to test the HAL / RAL
  - we don't have runnable examples in this project; instead, we rely on other projects to test the HAL. Provide links to other works that substantially depend on the HAL, like the `teensy4-bsp`.

Add other ideas below!

mciantyre:
Here's some ideas for the project's README. I also added a CONTRIBUTING guide for users who want to modify the project. These ideas came after discussing the current state of documentation (#51).

[Rendered README](https://github.com/mciantyre/imxrt-rs/blob/docs/README.md).

Closes #53.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,13 +8,13 @@ Thanks for helping us build embedded Rust support for NXP's i.MX RT processors! 
 
 ## Development
 
-The steps below are useful for developers who want to build the RAL and the HAL on their own systems. All steps below assume that you've cloned the repository.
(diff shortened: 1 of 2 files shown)


bfrog:
This is a great change, and unquestionably an improvement. Mergin.

PR #55, opened 2020-05-01
Add module-level, compiler-checked examples

mciantyre:
Keeping on my documentation kick, the PR adds module-level examples for almost all of the peripherals. `cargo test` will make sure that all of these examples compile. I derived them from the Teensy 4 examples.

I'm also including a fix to the HAL's `Cargo.toml`. The HAL's documentation [fails to build on docs.rs](https://docs.rs/crate/imxrt-hal/0.2.1/builds). We specify a feature-flag to use when generating the docs, which should fix the doc build. I tested it locally, but we'll know if it works the next time we release the HAL.

### DIFF
--- a/imxrt-hal/src/iomuxc.rs
+++ b/imxrt-hal/src/iomuxc.rs
@@ -1,34 +1,4 @@
 //! IOMUX Controller
-//!
-//! ## Configure pin settings
-//!
-//! This takes a given PinConfig and updates the *_PAD_MUX_PAD_*
-//! register related to a given Pin.
-//!
-//! If the PinConfig does not write to all fields, which is checked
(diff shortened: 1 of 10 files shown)


imxrt-hal/src/iomuxc.rs
```diff
@@ -1,4 +1,34 @@
 //! IOMUX Controller
+//!
+//! ## Configure pin settings
```
bfrog, line 3: This seems out of place to me, the module level docs here probably should encompass more of what iomuxc does as a whole rather than the PinConfig stuff

mciantyre: I can move into the `pin_config.rs` module-level docs, rather than the `iomxc.rs` docs.

PR #58, opened 2020-06-10
Support split UART peripherals

mciantyre:
I'm thinking about adding a new, UART-based logging back-end for the Teensy. Logging might only need the transfer half of a UART peripheral. This PR proposes split-able UART peripherals.

Users can `split()` a peripheral into transfer and receive halves. The transfer half only supports writes; the receive half, only reads. Users can `join()` the halves if they need to re-configure the
peripheral. The two halves implement the DMA's `Source` and `Destination` traits, so they may serve as uni-directional DMA sources and destinations.

Tested in the Teensy 4 repo, at [aca5424](https://github.com/mciantyre/teensy4-rs/commit/aca5424ea07cfe1421362cc7dd6a6912b5dad6a2).

### DIFF
--- a/imxrt-hal/CHANGELOG.md
+++ b/imxrt-hal/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - `uart::ReadError` implements `Clone`, `Debug`, `PartialEq`, and `Eq`
 - `UART` peripherals may be `split()` into `Tx` and `Rx` halves
+- `UART` peripherals implement the blocking `embedded_hal` traits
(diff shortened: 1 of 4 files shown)


bfrog:
I like the idea here, very nice!

mciantyre:
There's a proof-of-concept serial logger available in the [imxrt-uart-log](https://github.com/imxrt-rs/imxrt-uart-log) crate. The crate relies on split UART halves to log data over a serial interface.

Issue #15, opened 2020-03-02
Ignore generated ral code


bfrog:
I think it might make sense to ignore any generated code, to make it clearer that those files really shouldn't be edited by hand. It does mean any contributor would need to have python and some tooling installed (though already done semi-automatically via make venv) to generate the ral files.

I'd like some feedback on that before doing it.

bfrog:
I went ahead and did it, but will wait on merging based on feedback. I'm also looking to automatically get a CI flow going in github actions that generates the code and builds it at the very least.

PR #66, opened 2020-06-29
Check-in all RAL sources

mciantyre:
The PR adds all RAL sources to the repo. It undoes #15.

This PR is in draft util we identify a way to keep consistency with `imxrtral.py`. Whatever solution we choose will target this branch. We'll review the approach, and merge it into this PR. Then, we'll merge everything into `master`. This should make it easier for us to isolate the interesting changes from the RAL noise.


mciantyre:
See #67. I pushed to the wrong remote!

PR #70, opened 2020-07-05
Fix StatefulOutputPin and read from data register

mciantyre:
We should read PSR when the GPIO is an input pin. When its an output
pin, and we want to know what state we're in, we should instead read
the data register.

Found by @mitchmindtree during teensy4-rs prototyping. Mitch found
that is_set_high() always returned `false`, even when we could see
that the GPIO was high. I also reproduced this in the teensy4-rs'
USB example, printing the state of the LED. Reading the DR register
resolved the issue in my testing.


mciantyre:
See the bug report in mciantyre/teensy4-rs#67.

Holding as a draft until @mitchmindtree confirms this resolves the issue.

mitchmindtree:
Yep this fixes it!

Thanks for the speedy patch :)

PR #77, opened 2020-08-02
Support a safer IOMUXC interface

mciantyre:
When we first proposed the iomuxc crate (#73), we said that most of the API was `unsafe`. We stressed the unsafety of functions like `set_sion()`, `configure()`, and `prepare()`, since the functions acted on types that were implemented *outside* of the crate, and we couldn't attest to their correctness.

In #75, we moved to a single crate with feature-flags implementation. Now that all pads are implemented in the same crate, we can easily check correctness of the implementations. The PR marks the pin preparation and pad configuration & muxing functions as safe.

It it still unsafe to generate the pad objects. But, as long as an interface guarantees only one live instance of a pad, everything should be sound. Let me know if we see any issues, or if we should keep these `unsafe`.

(In hindsight, the `unsafe` traits are probably enough to signal unsafety, instead of also making the functions unsafe.)

### DIFF
--- a/imxrt-hal/src/gpio.rs
+++ b/imxrt-hal/src/gpio.rs
@@ -134,7 +134,7 @@ where
     ///
     /// All pads may be used as a GPIO, so this should always work.
     pub fn new(mut pin: P) -> Self {
-        unsafe { crate::iomuxc::gpio::prepare(&mut pin) };
+        crate::iomuxc::gpio::prepare(&mut pin);
         Self {
             pin,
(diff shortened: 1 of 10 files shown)


bfrog:
The safety makes sense to me, the API is safe so long as no one else is also doing unsafe things with pointers to the MMIO range, since we can verify there's no overlap in the crate itself (that's my assumption I'm making) then I think you are correct making these safe apis.

PR #81, opened 2020-09-06
IOMUXC prelude

mciantyre:
The PR adds a small `prelude` module to the `imxrt-iomuxc` crate. `prelude` re-exports the general types and modules from the crate. It's to help HAL developers simply re-export the `imxrt-iomuxc` crate API without exposing the processor-specific IOMUXC module.

Today's usage in the HAL

https://github.com/imxrt-rs/imxrt-rs/blob/bd1b2df5afae5694ba71f33bc85e8fd349e358e7/imxrt-hal/src/lib.rs#L16-L27

produces [this API](https://docs.rs/imxrt-hal/0.4.0/imxrt_hal/iomuxc/index.html). The first `pub use imxrt_iomuxc::*;` statement includes a re-export of the `imxrt106x` module. We then re-export that module's interface through `pub use imxrt_iomuxc::imxrt106x::*;`. The end result is that the processor-specific types are reachable through the HAL's `iomuxc` module, and also the HAL's `iomuxc::imxrt106x` module, which is redundant.

Since the prelude contains everything *except* the processor-specific module, HAL implementers can do

```rust
pub mod iomuxc {
    // Re-export common modules and types
    pub use imxrt_iomuxc::prelude::*;
    // Conditionally re-export chip-specific pads
    #[cfg(feature = "imxrt106x")]
    pub use imxrt_iomuxc::imxrt106x::*;
}
```

without exposing the processor-specific module.

### DIFF
--- a/imxrt-iomuxc/Cargo.toml
+++ b/imxrt-iomuxc/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "imxrt-iomuxc"
-version = "0.1.0"
+version = "0.1.1"
 authors = ["Ian McIntyre <ianpmcintyre@gmail.com>"]
 edition = "2018"
 description = """Pad configuration interface for NXP i.MX RT processors.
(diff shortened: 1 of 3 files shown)


mciantyre:
I put this together, but I'm still on the fence about it. My thinking was that it'd be helpful when considering a crate-per-HAL, since we can re-export the prelude and the processor-specific module without any module redundancies. But at the same time, the redundancy isn't that troublesome. Looking for other thoughts!

bfrog:
This makes sense to me, exporting a chip specific module seems like it could lead to trouble when trying to write things on top of multiple chips

PR #83, opened 2020-10-03
Implement basic support for the SRTC

AlyoshaVasilieva:
This implements very basic support for the real-time clock: enabling the SRTC and RTC, setting the SRTC, and getting the current time.

The implementation is a bit awkward, but I'm not sure how to improve it; unlike the other peripherals I know of, the SRTC keeps running until it loses power or it is specifically disabled, so having two methods of enabling it (one that sets the time, one that does not) makes sense. I guess a builder pattern (like I2C) could be used, but not sure if it's worth it currently.

Also, setting the time while the RTC is running also makes sense, so that needs to be available while the RTC is running (and requires disabling it).

This implementation also takes ownership of the Secure Non-Volatile Storage instance, but the SNVS does a lot more than just the (S)RTC. It also has:

* A monotonic counter
* Security key storage
* Power glitch detection
* General-purpose register (in the low-power domain, so may be preserved across reboot and through loss of main power)
* Probably other things

Taking ownership of the SNVS is therefore not ideal, since it's involved in those other things too. I've implemented reading the current time without the use of the SNVS instance (using an `unsafe` block) since I know that's safe, but I don't know what else is or isn't safe to do. I would like to improve this but don't know how.

This PR also adds the `.idea` directory to gitignore, since I use IDEA rather than VSCode.

### DIFF
--- a/imxrt-hal/src/srtc.rs
+++ b/imxrt-hal/src/srtc.rs
@@ -1,0 +1,144 @@
+//! (Secure) Real-Time Clock
+//!
+//! Basic support only: Supports enabling and setting the clocks, but not any of their more
+//! advanced features (alarm, interrupt, power glitch detection, etc).
+//!
+//! The Secure Real-Time Clock continues tracking time until it is specifically disabled or loses
+//! power, even through a system reboot (and potentially a loss of power to the system, if the
+//! SRTC is connected to a battery).
+//!
+//! # Example
+//!
+//! ```no_run
+//! use imxrt_hal;
+//!
+//! let mut peripherals = imxrt_hal::Peripherals::take().unwrap();
+//!
+//! let mut rtc = peripherals.rtc.enable_and_set(1600000000);
+//! // Interpreted as Unix time: Sep 13 2020 12:26:40
+//!
+//! let now = rtc.get();
+//! ```
+//!
+//! # Teensy 4.x Note
+//!
+//! When the SRTC is enabled, setting the board into program mode then using the Teensy Loader
+//! to reboot it will set the current time (Unix epoch, but time in local timezone). This will
+//! overwrite whatever time you may have previously set.
+
+use core::fmt;
+
+use crate::ral;
+use crate::ral::snvs::Instance;
+
+const MR_SHIFT: u8 = 17;
+const LR_SHIFT: u8 = 15;
+
+/// A disabled RTC.
+///
+/// Note: The SRTC may actually be enabled, as (unless disabled) it stays enabled across reboots
+/// as long as it has power.
+pub struct Unclocked {
+    reg: Instance,
+}
+
+impl Unclocked {
+    pub(crate) fn new(snvs: Instance) -> Self {
+        Unclocked { reg: snvs }
+    }
+
+    /// Enable both the Secure Real-Time Clock and the Real-Time Clock and set them to the
+    /// provided time. The time provided is
+    /// [a count of seconds since some epoch](https://en.wikipedia.org/wiki/Epoch_(computing)).
+    /// Usually that will be the Unix epoch, but as an example setting `0` would create a simple
+    /// 'seconds since RTC enabled' clock.
+    pub fn enable_and_set(mut self, time: u32) -> RTC {
+        disable(&mut self.reg);
+        set(&mut self.reg, time);
+        self.enable()
+        // a whole disable-set-enable cycle appears to take ~950us
+    }
+
+    /// Enable the Real-Time Clock without setting the time. The SRTC keeps track of time as long
+    /// as it is enabled and has power, so this function will avoid overwriting the old time.
+    ///
+    /// If no time had previously been set, the RTC will start counting from `0`.
+    pub fn enable(mut self) -> RTC {
+        enable(&mut self.reg);
+        RTC { reg: self.reg }
+    }
+}
+
+/// Disable the RTC and SRTC.
+fn disable(snvs: &mut Instance) {
+    // If the SRTC is locked this function will loop forever.
+    // SRTC locking is not implemented, so if it's locked then the user did it manually.
+    ral::modify_reg!(ral::snvs, snvs, HPCR, RTC_EN: RTC_EN_0); // disable RTC
+    ral::modify_reg!(ral::snvs, snvs, LPCR, SRTC_ENV: SRTC_ENV_0); // disable SRTC
+    while ral::read_reg!(ral::snvs, snvs, HPCR, RTC_EN) != 0
+        || ral::read_reg!(ral::snvs, snvs, LPCR, SRTC_ENV) != 0
+    {
+        core::sync::atomic::spin_loop_hint();
+    } // wait until RTC and SRTC turn off
+}
+
+/// Sets the SRTC time.
+fn set(snvs: &mut Instance, time: u32) {
+    ral::modify_reg!(ral::snvs, snvs, LPSRTCMR, SRTC: time >> MR_SHIFT);
+    ral::write_reg!(ral::snvs, snvs, LPSRTCLR, time << LR_SHIFT);
+    // The lowest 15 bits of MR are the MSB, upper are reserved. §20.6.6
+    // The upper 17 bits of LR are the LSB down to seconds.
+    // The remaining 15 bits of LR are slowly incrementing and appear to relate to the RTC's
+    // optional periodic interrupt. As we are changing the time, they are reset to 0.
+}
+
+/// Enable the RTC and SRTC, setting RTC to sync from SRTC.
+fn enable(snvs: &mut Instance) {
+    ral::modify_reg!(ral::snvs, snvs, LPCR, SRTC_ENV: SRTC_ENV_1); // enable SRTC
+    ral::modify_reg!(ral::snvs, snvs, HPCR, RTC_EN: RTC_EN_1, HP_TS: HP_TS_1); // enable RTC and sync
+    while ral::read_reg!(ral::snvs, snvs, HPCR, RTC_EN) == 0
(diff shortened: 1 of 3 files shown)


imxrt-hal/src/srtc.rs
```diff
@@ -0,0 +1,144 @@
+//! (Secure) Real-Time Clock
+//!
+//! Basic support only: Supports enabling and setting the clocks, but not any of their more
+//! advanced features (alarm, interrupt, power glitch detection, etc).
```
mciantyre, line 4: Does the power glitch detection apply to the RTC? Or, is it one of the other SNVS functions? If it's for the SNVS, we might remove it in this RTC documentation.
```suggestion
//! advanced features (alarm, interrupt, etc).
```

imxrt-hal/src/srtc.rs
```diff
@@ -0,0 +1,144 @@
...
+const MR_SHIFT: u8 = 17;
+const LR_SHIFT: u8 = 15;
+
+/// A disabled RTC.
+///
+/// Note: The SRTC may actually be enabled, as (unless disabled) it stays enabled across reboots
+/// as long as it has power.
+pub struct Unclocked {
+    reg: Instance,
+}
```
mciantyre, line 43: Just a note that the CCM clock gating register 5 (CCM_CCGR5) has two clock gate fields for the SNVS high- and low-power domains. The reset values are 'run'. According to "Clocks at Boot Time" section (table 9-7, iMXRT1060 reference manual, rev 2), the boot ROM does not disable these clocks.

We'll note that some other HAL drivers explicitly enable their clock gates, even if they are not disabled by boot ROM. We might consider doing the same here, but it would just be for consistency. We're OK without it!

mciantyre (commented):
Looking good to me! Tested on a Teensy 4. If you have a hardware example you'd like to contribute, I'm happy to favor yours over mine. Here's some thoughts to get our review started.

I like the initialization API, since it lets us either commit a starting time, or prefer whatever time the system is using. Do you think we could support an additional `try_enable` method that would only set the starting time if we detected the low-power domain is enabled? This might satisfy the use case "I want to set the baseline time, but *only* if there isn't an existing baseline time." Pseudo-code below.

```rust
/// Indicates the result of the `try_enable` method
pub enum EnabledState {
    /// The RTC is already enabled, and it's currently counting from `seconds`
    AlreadyCounting {
      rtc: RTC,
      /// The current time on the RTC
      seconds: u32
    },
    /// The RTC was not previously enabled, and it's now counting from the
    /// `seconds` supplied to `try_enable`
    SetTime(RTC),
}

impl Unclocked {
    /// Enable the RTC
    ///
    /// If the RTC isn't already running, `try_enable` uses `seconds` as the starting count for the RTC.
    /// If the RTC is currently running, the return indicates the current RTC time. 
    pub fn try_enable(self, seconds: u32) -> EnabledState {
        // ...
    }
}
```

Taking the `SNVS` instance is OK with me. As long as it's marked as taken by *something*, it will signal to the user that they shouldn't also take it. Future drivers that also need to use `SNVS` can use a stolen instance, and we'll make sure things are safe in the HAL crate.

Alternatively, we could refactor the design, and implement a SNVS HAL driver, instead of a dedicated RTC driver. Today's SNVS will only has RTC methods, and we can later expand it to support future SNVS functionality. Since it might be non-obvious to users, we teach them through documentation and examples. It might burden on the end user, since they would need to figure out how to share the SNVS for their independent RTC and secure key storage needs.

imxrt-hal/src/srtc.rs
```diff
@@ -0,0 +1,144 @@
...
+//! // Interpreted as Unix time: Sep 13 2020 12:26:40
+//!
+//! let now = rtc.get();
+//! ```
+//!
+//! # Teensy 4.x Note
+//!
+//! When the SRTC is enabled, setting the board into program mode then using the Teensy Loader
+//! to reboot it will set the current time (Unix epoch, but time in local timezone). This will
+//! overwrite whatever time you may have previously set.
```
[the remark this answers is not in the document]
AlyoshaVasilieva: It appears to specifically be the graphical application, yes. I initially wrote that note as I was very confused before figuring it out; it's almost totally undocumented and isn't mentioned by the Teensy Loader GUI even in the verbose info box.

AlyoshaVasilieva:
While implementing `try_enable` I realized that it would have to read from the SRTC to get the current time, since the RTC is not preserved across reboots. While doing so, I realized that the SRTC's `LPSRTCLR` register contains the ticks of the 32768Hz clock instead of some interrupt-related value. That allows for sub-second times to be retrieved and set, not just whole second. And that made me realize there's no point in using both the SRTC and RTC in one struct.

So: this now only enables and uses the SRTC, it ensures the SNVS_LP clock gate is enabled, and it supports setting and retrieving the ticks (plus convenience functions for ms/tick conversions; less accurate than I'd prefer as no_std doesn't support `round`).

It's slightly more complex to use but users can just ignore the ticks by setting 0 and using `get` rather than `get_with_ticks`. Ticks aren't auto-converted to/from milliseconds (or microseconds) to avoid everyone needing to pay for that conversion.

imxrt-hal/src/srtc.rs
AlyoshaVasilieva: Added clock gate enable using the CCM handle.

AlyoshaVasilieva:
I'll look into writing a hardware example that includes time setting. After experimenting, I think the Teensy Loader GUI's RTC auto-set is useless for us, as local-seconds-since-Unix-epoch is ambiguous around daylight savings.

bfrog (approved):

bfrog:
I might change get_f64 into a split of whole seconds and microseconds and avoid the floats

AlyoshaVasilieva:
Switched to microseconds. I don't know how to write this entirely without floats but I did eliminate the divide.

imxrt-hal/src/srtc.rs
```diff
@@ -117,27 +118,29 @@ fn is_enabled(snvs: &Instance) -> bool {
...
-    let millis = millis.min(999);
-    let millis = millis as f32 / 1000.0;
-    let ticks = millis * 32768.0;
+/// For example: for the time `2020-10-05 01:39:56.505`, `micros` is `505000`, and this gives
+/// `16547` ticks.
+pub fn subsec_micros_to_ticks(micros: u32) -> u16 {
+    const RATIO: f32 = 0.032768; // 32768 / 1_000_000, each microsecond is this many ticks
+    let micros = micros.min(999_999);
+    // 999999 is the max valid value outside of a leap second; clamp
+    let ticks = micros as f32 * RATIO;
```
bfrog, line 130: So doing integer math usually what you want to do is scale to some more precise unit first, do math, then scale back to the less precise unit. So for example.... 1_000_000_000/32_768 gets you the number of nanoseconds per tick (30517) as an integer. Convert the microseconds to nano seconds (multiply by 1000), divide by nanoseconds per tick, get the number of ticks.

So in real terms.... 
``` rust
const NANOS_PER_TICK: u32 = 30517;
fn microseconds_to_ticks(micros: u32) -> u16 { 
  let nanos = micros *1000;
  nanos/NANOS_PER_TICK
}
```

bfrog: I'm not sure this function needs to be public, if it is lets avoid the micros.min() and call it micros_to_ticks

imxrt-hal/src/srtc.rs
```diff
@@ -117,27 +118,29 @@ fn is_enabled(snvs: &Instance) -> bool {
...
 
 // these won't round trip in part due to the lack of round in no_std
 
-/// Converts clock ticks (1/32768 of a second) into milliseconds.
+/// Converts sub-second clock ticks (1/32768 of a second) into microseconds.
 ///
-/// For example: 32000 ticks works out to 976 milliseconds.
-pub fn ticks_to_subsec_millis(ticks: u16) -> u16 {
+/// For example: 32000 ticks works out to 976562 microseconds.
+pub fn ticks_to_subsec_micros(ticks: u16) -> u32 {
```
bfrog, line 139: The same math can be done above but with (ticks*NANOS_PER_TICK)/1000 to get microseconds

bfrog: The same commentary about making it public, if its public I think ticks_to_micros is good enough, that's what the math is doing here

imxrt-hal/src/srtc.rs
```diff
@@ -211,6 +214,13 @@ impl SRTC {
         get(&self.reg)
     }
 
+    /// Gets the current time as a tuple containing the count of seconds since some point in the past
+    /// and the sub-second time as microseconds.
+    pub fn get_with_micros(&self) -> (u32, u32) {
```
bfrog, line 219: Great!

bfrog (changes requested):

imxrt-hal/src/srtc.rs
AlyoshaVasilieva: Thanks!

I've changed the code to use quarter-nanoseconds per tick, which is slightly more accurate than just nanoseconds but still fits within 32-bit unsigned math.

`micros_to_ticks` needs to be public as it is not used internally, it's for users to convert sub-second time into clock ticks so they can more easily set the time.

I've removed the `min` call and documented the max valid value instead.

imxrt-hal/src/srtc.rs
AlyoshaVasilieva: Changed the math and made it private, which also led me to remove the 15-bit mask as we now know the input is always going to be 15-bit since it comes directly from the clock.

bfrog (approved):
I think this looked good

mciantyre:
LGTM! Thanks again for the new timer.

Issue #20, opened 2020-03-03
Configure/Release


teburd:
Configuring a peripheral should be undoable in some way to return it back to its default state.

The C SDK does this using _init/_deinit pairs for peripherals, we can probably do this using type states and similar calls, instead perhaps using configure/release as the call pair.

bfrog:
Really we are doing clock, so perhaps unclock is the right term to use here anyways

bfrog:
There's really a few possible States a peripheral may be in beyond clocked, but the goal is the same, a reversal of the peripheral states from Unclocked -> Clocked -> Configured and back

PR #84, opened 2020-10-07
ADC Implementation

### LINKED ISSUE #47: ADC implementation
`embedded-hal` specifies ADC traits [here](https://docs.rs/embedded-hal/0.2.3/embedded_hal/adc/index.html)

It seems that this crate does not implement those yet. This would be useful in order to e.g. read analog values from the pins of a Teensy 4 board.

---

_Would love to try and help with this, if someone could point me to the right direction & provide a bit of mentoring 🙂_

DavidTheFighter:
This PR adds an ADC functionality and implements the `embedded-hal` 0.2.3 traits (0.2.4 makes a few small changes). It closes #47. 
This has all the features that I will reasonably use, and has a couple extra features for power users. I decided to go with a simpler interface rather than letting the user change every setting possible, so real power users might want more functionality but we can always add that later. This was tested on a Teensy 4.1
Feel free to nitpick over things :)


bfrog:
Looks like perhaps some doc comments don't build correctly, will look closer when this passes all the checks, super excited to get this!

DavidTheFighter:
There we go. Sorry that was a bad combination of me not realizing there was CI checks and being distracted by a class.

imxrt-hal/src/adc.rs
```diff
@@ -0,0 +1,275 @@
...
+{
+    /// Creates a new analog input pin
+    pub fn new(mut pin: P, _adc: &ADC<ADCx>) -> Self {
+        prepare_adc_pin(&mut pin);
+        Self {
+            _module: PhantomData,
+            _pin: pin,
+        }
+    }
+}
```
mciantyre, line 109: Should we add a `release` method to re-acquire the pad? Seems fitting with #20.
```suggestion
    }
    
    /// Release the ADC input, returning the underlying hardware pin
    pub fn release(self) -> P {
        self.pin
    }
}
```

imxrt-hal/Cargo.toml
```diff
@@ -45,4 +45,4 @@ rt = ["imxrt-ral/rt"]
 nosync = ["imxrt-ral/nosync"]
 
 [package.metadata.docs.rs]
-features = ["imxrt1062"]
```
bfrog, line 48: It's unclear why this change is here

imxrt-hal/src/adc.rs
```diff
@@ -0,0 +1,275 @@
...
+//! use imxrt_hal::{self, adc};
+//! use embedded_hal::adc::OneShot;
+//!
+//! let mut peripherals = imxrt_hal::Peripherals::take().unwrap();
+//! let (adc1_builder, _) = peripherals.adc.clock(&mut peripherals.ccm.handle);
+//!
+//! let mut adc1 = adc1_builder.build(adc::ClockSelect::default(), adc::ClockDivision::default());
+//! let mut a1 = adc::AnalogInput::new(peripherals.iomuxc.ad_b1.p02, &adc1);
+//!
+//! let reading: u16 = adc1.read(&mut a1).unwrap();
```
bfrog, line 16: Very nice

imxrt-hal/src/adc.rs
```diff
@@ -0,0 +1,275 @@
...
+    }
+}
+
+impl<P, ADCx> AnalogInput<ADCx, P>
+where
+    P: Pin<ADCx>,
+    ADCx: adc::ADC,
+{
+    /// Creates a new analog input pin
+    pub fn new(mut pin: P, _adc: &ADC<ADCx>) -> Self {
```
bfrog, line 102: It might be more interesting to obtain an AnalogInput from the ADC

So rather than AnalogInput::new(...)

You'd get something like...

``` rust
let adc1_channel = adc1.channel(pin);
```

This could be done by creating a quick trait and impl per channel perhaps. Something about passing the reference to the ADC here seems off putting to me, I was thinking it might be shared among all AnalogInputs but that's clearly not the case!

mciantyre (commented):
Looks great! Thanks for putting this together.

Haven't yet tested on hardware. If you have a larger example that you're share, we'd be happy to accept it here, or in the `teensy4-rs` repo. Otherwise, I may put one together this weekend.

I'm not reviewing some details, such as making sure match arms like `AveragingCount::Avg32 => AVGS_3` are the correct mapping. Let me know if you'd like another set of eyes on that.

---

Thoughts on documentation: we're putting the `ADC` into a default state:

- 10 bits
- hight power
- slow conversion
- etc...

should we document that's the default state, so that users can rely on it? (I see some comments that hint at the default hardware state; is that correct?) Or, should we put documentation resembling "the default state is undefined; you must use `set_resolution` and [the other methods] to specify your state"?

imxrt-hal/src/adc.rs
```diff
@@ -0,0 +1,275 @@
...
+        // Calibrate w/ slow settings initially
+        inst.set_averaging(AveragingCount::Avg32);
+        inst.set_conversion_speed(ConversionSpeed::Slow);
+        inst.calibrate();
+
+        // Set to default of 4 hardware averages & medium conversion speed
+        inst.set_averaging(AveragingCount::Avg4);
+        inst.set_conversion_speed(ConversionSpeed::Medium);
+
+        inst
```
bfrog, line 135: I think the configuration functions above make sense, I do have a concern about them once interrupts and some sort of streaming comes about. What happens if you tell the ADC to stream conversions and then change the settings while its streaming? Would it make more sense to provide a Configuration struct and let the ADC then configure itself as needed entirely? Would it save any register modify calls? Would it make changing the configuration in an atomic (from the callers perspective) easier? Something to maybe consider

imxrt-hal/src/adc.rs
```diff
@@ -0,0 +1,275 @@
...
+    ADCx: adc::ADC,
+    WORD: From<u16>,
+    P: Pin<ADCx>,
+{
+    type Error = ();
+
+    /// Read an ADC value from an AnalogInput. This should always return a good result
+    fn read(&mut self, _pin: &mut AnalogInput<ADCx, P>) -> nb::Result<WORD, Self::Error> {
+        let channel = <P as Pin<ADCx>>::Input::U32;
+        ral::modify_reg!(ral::adc, self.reg, HC0, |_| channel);
```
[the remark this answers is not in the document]
DavidTheFighter: Fair point!

imxrt-hal/src/adc.rs
DavidTheFighter: Sounds good to me. How should we "unprepare" the pin? An `unprepare` function in `imxrt-iomuxc`?

imxrt-hal/Cargo.toml
DavidTheFighter: I don't either, by the looks of it a stray newline got deleted. My bad!

imxrt-hal/src/adc.rs
DavidTheFighter: I definitely agree that's a janky. I tried a couple things but I'm new enough to Rust I couldn't get the generics to work out. I'll take a look at it!

imxrt-hal/src/adc.rs
DavidTheFighter: Fair point. While it wouldn't save any register modify calls, I'm fine with making a configuration struct. I'm not sure what you mean about streaming conversion, though. Do you mean putting an ADC in a continuous mode? I decided not to do continuous mode for the time being because `embedded-hal` has no traits for it. But in that case I feel like disabling any configuration changes while streaming is the easiest way of going about it, however that be implemented. Otherwise, for the one shot reads, I think putting them in an interrupt free block would be the easiest way there (assuming there aren't any implications with that, I'm not familiar with that side of the code).

DavidTheFighter:
I'm for putting a guaranteed default state in the docs. Both to make it easier for new users and to make it more clear how the ADC is configured when it's created.

imxrt-hal/src/adc.rs
```diff
@@ -0,0 +1,275 @@
...
+use crate::iomuxc::{adc, consts::Unsigned};
+use crate::ral;
+use core::marker::PhantomData;
+use embedded_hal::adc::{Channel, OneShot};
+
+/// The clock input for an ADC
+#[allow(non_camel_case_types)]
+pub enum ClockSelect {
+    IPG,   // IPG clock
+    IPG_2, // IPG clock / 2
```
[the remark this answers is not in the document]
DavidTheFighter: According to the docs, the ADC does have an ideal frequency range (with tradeoffs still, eg speed vs accuracy), but I couldn't find any actual numbers on this besides minimum and maximum frequencies in the electrical characteristics for the ADC in the datasheet (which I'm assuming has nothing to do with that "ideal" range). Beyond that there's nothing I'm aware of that requires knowing the IPG frequency. To be honest that was one the big reason I made the default `ClockSelect` the asynchronous clock (ADACK), and the division and HSC mode to what they are is the Teensy core libraries use that configuration. I'm fine with accepting an IPGFrequency for the time being, if anything to make a more uniform interface across the hal.

imxrt-hal/src/adc.rs
```diff
@@ -0,0 +1,275 @@
...
+}
+
+/// Implement embedded-hal traits
+impl<ADCx, WORD, P> OneShot<ADCx, WORD, AnalogInput<ADCx, P>> for ADC<ADCx>
+where
+    ADCx: adc::ADC,
+    WORD: From<u16>,
+    P: Pin<ADCx>,
+{
+    type Error = ();
```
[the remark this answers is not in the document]
DavidTheFighter: I'm ambivalent either way, so I'll take your suggestion to use `core::convert::Infallible`.

imxrt-hal/src/adc.rs
```diff
@@ -0,0 +1,275 @@
...
+use crate::ccm;
+use crate::iomuxc::adc::{prepare_adc_pin, Pin, ADC1, ADC2};
+use crate::iomuxc::{adc, consts::Unsigned};
+use crate::ral;
+use core::marker::PhantomData;
+use embedded_hal::adc::{Channel, OneShot};
+
+/// The clock input for an ADC
+#[allow(non_camel_case_types)]
+pub enum ClockSelect {
```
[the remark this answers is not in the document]
DavidTheFighter: Fine by me!

imxrt-hal/src/adc.rs
DavidTheFighter: Whoa, brain fart. Not entirely sure why I was passing in the ADC, probably something about generics that were screwing with me while I was figuring them out. Does changing it to `let mut a1 = adc::AnalogInput::new(peripherals.iomuxc.ad_b1.p02);` look ok?

bfrog: that seems good to me

imxrt-hal/src/adc.rs
bfrog: As long as the only method for using the ADC is a polling method where the code flow is blocked, this is fine and safe and friendly.  In an interrupt driven conversion you really want to change the config of the ADC in a somewhat atomic way, either stopping the ADC, changing config, starting it again, or dropping values you know don't care about while the config is changing. You don't want your desired ADC config to be up in the air as to what it actually is. That can be done by using a configure(cfg: AdcConfiguration) type function for people, or it can be done by users themselves however they like. It's something to keep in mind. I don't think one way or the other is bad, the functions you have are needed regardless and we can add on an atomic-like configure function later.

imxrt-hal/src/adc.rs
mciantyre: Good thought! There hasn't been a need for an `unprepare` function yet. We could unprepare to the reset state, but that would require us to specify the reset states in code...

We can release the pad in an unspecified state. If a user wants to re-use the pad for something else, they will call a different `prepare` function on the pad.

imxrt-hal/src/adc.rs
```diff
@@ -0,0 +1,275 @@
...
+
+        // Set to default of 4 hardware averages & medium conversion speed
+        inst.set_averaging(AveragingCount::Avg4);
+        inst.set_conversion_speed(ConversionSpeed::Medium);
+
+        inst
+    }
+
+    /// Sets the resolution that analog reads return, in bits
+    pub fn set_resolution(&self, bits: ResolutionBits) {
```
mciantyre, line 139: This kind of pertains to the above conversation about interrupt safety: should these `set_*` methods take a `&mut self` receiver, rather than a `&self`? Same for `calibrate()`? They're mutating something owned by the ADC.

The `Oneshot` method already takes a `&mut self` receiver. Once everything it signaling mutability, anyone who wants to modify an `ADC` across execution contexts will need to wrap it in a mutex, or their favorite interrupt-free construct, to synchronize it across execution contexts.

bfrog: I think that mostly solves my concerns if you want to expose the set_ style configuration functions, they should take a &mut self, good call.

It sort of leaves it up to the user to figure out how to synchronize and make it safe. In the future perhaps a configure(&self, cfg: Configuration) could be provided which does either a disable, config, enable or an atomic spin loop waiting for interrupts to stop doing things perhaps.

DavidTheFighter:
I implemented all of your guy's suggestions, how's it look? @bfrog does changing the set_style methods to take a mutable reference to self settle your concerns?

mciantyre (approved):
Still looks great to me!

If there's no other work or review thoughts, I'll append a CHANGELOG entry, and merge this in a few days.

bfrog (approved):
Looks great to me, require mut in the places here enforces the single ownership requirements, perfect.

PR #85, opened 2020-10-09
Imxrt101x iomuxc

bfrog:
Adds the basic uart pin defs and beginnings of the imxrt101x iomux stuff to the repo

I ran into an annoyance in that the build (cargo build) in the project root fails in imxrt-hal if I don't specify imxrt1062 as default feature, I need to check the master branch again here first if that's how it was (been a bit since I did stuff!)

It's a nice feeling when you can git clone and cargo build though so maybe there's a nicer way of doing this or maybe this won't be an issue when I split the imxrt-hal stuff up more

### DIFF
--- a/imxrt-iomuxc/src/imxrt101x/mod.rs
+++ b/imxrt-iomuxc/src/imxrt101x/mod.rs
@@ -1,0 +1,100 @@
+//! Pads for the i.MX RT 101x processor family
+//!
+//! The module exports all of the i.MX RT 101x processor's pads. Pads that can support peripheral
+//! functions are tagged with `imxrt-iomuxc` traits.
+//!
+//! # Example
+//!
+//! In the example below, we implement a hypothetical `uart_new` function, which is responsible
+//! for preparing a UART peripheral. To properly configure the peripheral, we need the two
+//! pads that represent a peripheral's TX and RX pins. The implementation will use the
+//! `imxrt_iomuxc::uart::prepare()` function to prepare the pins.
+//!
+//! Note the trait bounds on `uart_new`. The usage requires that
+//!
+//! - the user provides one TX and one RX pin
+//! - the modules for each pin match
+//!
+//! ```no_run
+//! use imxrt_iomuxc as iomuxc;
+//! use iomuxc::uart::{Pin, TX, RX};
+//!
+//! # struct UART;
+//! /// Creates a UART peripheral from the TX and RX pads, and a baud rate
+//! fn uart_new<T, R>(mut tx: T, mut rx: R, baud: u32) -> UART
+//! where
+//!     T: Pin<Direction = TX>,
+//!     R: Pin<Direction = RX, Module = <T as Pin>::Module>,
+//! {
+//!     // Check the imxrt-iomuxc documentation to understand why
+//!     // this is unsafe.
+//!     unsafe {
+//!         iomuxc::uart::prepare(&mut tx);
+//!         iomuxc::uart::prepare(&mut rx);
+//!     }
+//!     // Prepare the rest of the UART peripheral, and return it...
+//!     # UART
+//! }
+//!
+//! # let sd_12 = unsafe { imxrt_iomuxc::imxrt101x::sd::SD_12::new() };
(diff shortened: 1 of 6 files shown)


imxrt-iomuxc/src/imxrt101x/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+//! #     unsafe {
+//! #         iomuxc::uart::prepare(&mut tx);
+//! #         iomuxc::uart::prepare(&mut rx);
+//! #     }
+//! #     UART
+//! # }
+//! # let gpio_09 = unsafe { imxrt_iomuxc::imxrt101x::gpio::GPIO_09::new() };
+//! # let gpio_10 = unsafe { imxrt_iomuxc::imxrt101x::gpio::GPIO_10::new() };
+//! // GPIO_10 is a UART1 TX pin, and GPIO_09 is a UART1 RX pin
+//! uart_new(gpio_10, gpio_09, 115_200);
```
mciantyre, line 88: This documentation test was designed to fail. It should show that supplying pads for separate UART peripherals does not compile.

https://github.com/imxrt-rs/imxrt-rs/blob/017a970596a59145dce29342fb2a6e45f2fa5722/imxrt-iomuxc/src/imxrt106x/mod.rs#L85-L88

It looks like these are two UART1. Consider mismatching the UART pins.

mciantyre (commented):
Looks good to me!

I've not studied the reference manual to ensure that the addresses, pad names, or UART pad assignments are correct. Let me know if you'd like that level of review.

Only concern is the new default feature, and how it might affect our releases and documentation.

> the build (cargo build) in the project root fails in imxrt-hal if I don't specify imxrt1062 as default feature

That's a limitation of our virtual workspace structure. Virtual workspaces don't support feature flags. Our [CONTRIBUTING guide](https://github.com/imxrt-rs/imxrt-rs/blob/017a970596a59145dce29342fb2a6e45f2fa5722/CONTRIBUTING.md#hal) recommends that we `cd imxrt-hal`, then build with feature flags, to work around that limitation. Specifying the manifest path from the workspace root, like

```
cargo build --features imxrt1062 --manifest-path imxrt-hal/Cargo.toml
```

 also works.

An alternative structure would eliminate the virtual workspace, and make `imxrt-hal` the core package. Then, we add a (non-virtual) workspace to include the `imxrt-ral` and `imxrt-iomuxc` crates. It would basically look like

```bash
$ mv imxrt-hal/* .
$ rm -Rf imxrt-hal
$ cat >> Cargo.toml <<EOF
heredoc> [workspace]
heredoc> members = [
heredoc>     "imxrt-ral",
heredoc>     "imxrt-iomuxc",
heredoc> ]
heredoc> EOF
$ vim Cargo.toml # Fix path dependencies
$ cargo build --features imxrt1062
$ vim .github/workflows/rust.yml # Fix CI
```

This supports feature specification at the repo's root, while putting the HAL front-and-center.

imxrt-hal/Cargo.toml
```diff
@@ -28,7 +28,7 @@ path = "../imxrt-iomuxc"
 bench = false
 
 [features]
-default = ["embedded-hal/unproven"] # Allows us to access the new digital pin traits
+default = ["embedded-hal/unproven", "imxrt1062"] # Allows us to access the new digital pin traits
```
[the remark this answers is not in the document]
bfrog: I plan on removing this default feature flag, the reasoning above is good, and in the future I think it'll get fixed in a nicer way as I start splitting things up as described in #56

imxrt-iomuxc/src/imxrt101x/mod.rs
bfrog: 👍

bfrog:
I need to experiment more I think with how #56 will play out with the workspace and such. We may split the repo, or change the shape of it like suggested above.

I think ideally though what happens instead is imxrt-ral changes to always provide common register block types that a common imxrt-hal can use, regardless of the chip being built for. Only chip specific register block types would then be brought in, for each imxrt101x-hal, imxrt106x-hal kind of crate perhaps. 

Something I need to tinker with as I move this forward.

PR #86, opened 2020-10-13
DMA support for all chip variants

mciantyre:
The PR provides DMA support for all chip variants. DMA support is probably lower on #56's TODO list. However, the DMA implementation is more unique than the other peripherals: the RAL interface was hand-written. Hand-writing the RAL was necessary for providing a simpler API, since the `svd2ral` script generated a lot of symbols.

I arrived at this implementation by studying the reference manuals. In summary, the 1011 chip only has 16 DMA channels; chips that are 1015 and higher hav 32 DMA channels. When you have more than 16 DMA channels, you have the option for channel grouping. Today's DMA driver doesn't support channel grouping, so this is only a comment in the code.

The conditional compiles are hidden behind a new `chip` module, private to the DMA driver. We use the common symbols from the `chip` module throughout the implementation.

This is a backwards-compatible change, at least in the short term. We now export a new constant, `CHANNEL_COUNT`, which signals the number of DMA channels supported in the implementation. Crate users should favor that constant instead of hard-coded `32`s.

### DIFF
--- a/imxrt-hal/src/dma.rs
+++ b/imxrt-hal/src/dma.rs
@@ -474,7 +474,7 @@ impl Unclocked {
     /// Enable the clocks for the DMA peripheral
     ///
     /// The return is an array of 32 channels. However, **only the first [`CHANNEL_COUNT`](constant.CHANNEL_COUNT.html) channels
-    /// are initialized to `Some(channel)`. The rest are `None`.
+    /// are initialized to `Some(channel)`. The rest are `None`.**
     ///
     /// Users may take channels as needed. The index in the array maps to the DMA channel number.
     pub fn clock(mut self, ccm: &mut ccm::Handle) -> [Option<Channel>; 32] {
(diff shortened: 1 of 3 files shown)


mciantyre:
Draft until we figure out

- [ ] what `DMA_CHANNEL_COUNT` should be if there are no feature flags
- [ ] what to do when users do something like this in their "generic" code:

```rust
let mut dma_channels = peripherals.dma.clock(&mut peripherals.ccm.handle);

let tx_channel = dma_channels[9].take().unwrap();  // OK; 9 < 16
let mut rx_channel = dma_channels[25].take().unwrap(); // Err, 25 > 16, won't work for 1011 generic code
```

bfrog:
So the above makes me think channel_count_32 should be a feature flag enabled by default, and when enabled you don't have to worry about the None type for a channel. I can disable the default feature for 1011 in the hal

Issue #87, opened 2020-10-15
IOMUXC: full support for 101x


mciantyre:
This issue tracks the `imxrt-iomuxc` traits implementations for the 101x family. See imxrt-rs/imxrt-async-hal#4 for more details. The **bolded** peripherals are required for fully using today's async HAL.

- [ ] ADC
- [x] **GPIO**
- [ ] **I2C**
- [ ] PWM
- [ ] **SPI**
- [x] **UART**

mciantyre:
#85 introduced the `imxrt101x` IOMUXC feature, with UART and GPIO support.

PR #88, opened 2020-10-15
IOMUXC: Add I2C, SPI traits for 1011 pads

mciantyre:
The PR adds I2C and SPI `imxrt-iomuxc` trait implementations for the 1011 pads.

Part of #87.


bfrog:
Thanks for doing this, will verify against the datasheet sometime today

bfrog (approved):
I verified the pins against the ref manual, and it looks good to me

PR #89, opened 2020-10-16
Removes conditional feature compilation for HAL

bfrog:
To start the hal split move the entire imxrt-hal crate to imxrt1062-hal,
set all feature flags for dependencies appropriately, and fix the
workspace to build on cargo build after checkout.

A second imxrt1011-hal can now easily be added, and shared code moved
appropriately to a common imxrt-hal-common crate as a dependency of
both as time and energy permits.


imxrt1062-hal/src/ccm.rs
```diff
@@ -925,30 +921,6 @@ pub mod spi {
...
-        LPSPI_PODF_12 = ccm::CBCMR::LPSPI_PODF::RW::LPSPI_PODF_12,
-        /// 0b1101: divide by 14
-        #[cfg(features = "imxrt1011")]
-        LPSPI_PODF_13 = ccm::CBCMR::LPSPI_PODF::RW::LPSPI_PODF_13,
-        /// 0b1110: divide by 15
-        #[cfg(features = "imxrt1011")]
-        LPSPI_PODF_14 = ccm::CBCMR::LPSPI_PODF::RW::LPSPI_PODF_14,
-        /// 0b1111: divide by 16
-        #[cfg(features = "imxrt1011")]
-        LPSPI_PODF_15 = ccm::CBCMR::LPSPI_PODF::RW::LPSPI_PODF_15,
```
mciantyre, line 951: OK with me to remove these. But, if we choose to keep them, it might make it easier to move this enum to a common HAL without having to remember that there are more variants for a 1011.

mciantyre (approved):
LGTM!

Before merging, can we update the [CONTRIBUTING](https://github.com/imxrt-rs/imxrt-rs/blob/master/CONTRIBUTING.md) guide to describe this change? This PR affects what's described in the HAL section.

Issue #94, opened 2020-10-24
RFC: Maintain i.MX RT-specific drivers in separate repos


mciantyre:
In #56, we discuss a split HAL, which will have a single HAL crate per i.MX RT processor family. The HALs will probably depend on `imxrt-iomuxc`, our custom driver for pad multiplexing and configuration.

We maintain `imxrt-iomuxc` in this repository. However, this issue proposes that we **maintain `imxrt-iomuxc`, along with other foundational, chip-specific drivers, in separate repositories**. The approach will have us maintain only the split HALs, the common HAL, and the RAL in this repository.

I define a _"foundational, chip-specific driver"_ as a library that 

- is specialized for i.MX RT processors
- could be used across higher-level i.MX RT libraries within our organization, and outside our organization
- fully replaces the equivalent API provided by `imxrt-ral`, so that the user does not require the RAL
- requires a higher-level, safe interface before exposing to end-users

Drivers that meet this criteria include

- IOMUXC, released as `imxrt-iomuxc`
  - API for pad configuration and multiplexing, specific for our processors
  - Used in HAL, [async HAL](https://github.com/imxrt-rs/imxrt-async-hal); depended on by [`teensy4-pins`](https://crates.io/crates/teensy4-pins)
  - Does not depend on the RAL
- CCM, in development as `imxrt-ccm`
  - API for clock configuration, and clock gating. Prototyped in `ccm` module of async HAL, but considered for HAL integration in #92
  - By default, usable without the RAL. Will _optionally_ depend on RAL as a convenience for RAL-based APIs (#92).
- DMA, in development as `imxrt-dma`
  - Unsafe API for allocating DMA channels, and scheduling DMA transfers. Re-exported by HAL and async HAL behind a safe interface.
  - Developed in HAL, then ported to async HAL with minimal changes. There's much code and documentation that could be shared.
  - Will not depend on the RAL

I do not consider HAL crates in the category of foundational, chip-specific drivers. HALs are

- designed for safety and ergonomics. Foundational, chip-specific drivers may not provide a safe interface, or may require a deeper understanding of the peripheral.
- intended to abstract processor complexities behind interfaces, like `embedded-hal`. Foundational, chip-specific drivers expect you to deal with that complexity.
- dependent on foundational, chip-specific drivers, and may re-expose the interfaces of these drivers. Foundational, chip-specific drivers may be used by themselves, possibly with no required dependencies.

## Pros

[...]

PR #96, opened 2020-10-25
Remove imxrt-iomuxc from the repository

mciantyre:
The PR removes the `imxrt-iomuxc` crate from this repository. I've migrated the package into a separate repository, [imxrt-iomuxc](https://github.com/imxrt-rs/imxrt-iomuxc). We may continue adding pin definitions, configurations, and multiplexing features in that repository.

If you have pending changes to `imxrt-iomuxc`, you should apply them onto the separate repository.

See #94 for more information.


bfrog (approved):
LGTM

PR #97, opened 2020-11-07
Fork DMA module into imxrt-dma

mciantyre:
The PR forks the HAL's `dma` module into a separate crate, `imxrt-dma`, then removes the crate from the HAL. [`imxrt-dma`](https://github.com/imxrt-rs/imxrt-dma) provides a lower-level API for coordinating DMA transfers on i.MX RT processors. I prototyped the interface as part of the [async HAL](https://github.com/imxrt-rs/imxrt-async-hal) development, then integrated it back into the HAL. The approach lets us share the core DMA types across our HALs.

The HAL extends `imxrt-dma` with the static buffer interface, peripheral abstractions, and the memcpy helper. These components are specific to the HAL, and it shows how a HAL designer can build a safer interface atop `imxrt-dma`. The HAL's DMA API remains backwards compatible.

You'll find that all commits before 4d480a0 are mirrored in the `imxrt-dma` repository. Unfortunately, `git subtree` isn't smart enough to track all the file movements which occurred up to this point (`imxrt-hal => imxrt1062-hal => imxrt1060-hal`, split into `imxrt-dma`). So, `imxrt-dma` doesn't track all of the DMA development history. I've taken imxrt-rs/imxrt-dma#1 to manually collect relevant PR and changes to the HAL's DMA module, so we can keep track of the history.

The useful parts of `imxrt-dma`, which used to be private to the DMA module, are qualified as unsafe. If there are any concerns about safety, can discuss that in the `imxrt-dma` issue tracker.

Tested in the async HAL crate, and in the `teensy4-bsp` examples. Part of #94.


bfrog (approved):
Looks good to me, and makes sense!

PR #100, opened 2020-11-15
Only allow HAL to create I2C/SPI config errors

AlyoshaVasilieva:
These errors are only valid if the HAL is creating them, so don't allow users to create them. Pretty minor.


mciantyre:
I probably set the bad precedent. Thank you for cleaning these up!

PR #102, opened 2020-11-17
Fix comment, don't preserve bad value in TRNG disable

AlyoshaVasilieva:
Woke up today and almost immediately realized that `0` is not inside the range `1..=15`. Woops.

I doubt this code change will actually matter, since users can't set a 0 retry count, but I didn't want to leave an incorrect comment laying around.


mciantyre:
Great catch, thanks for the fix!

PR #106, opened 2021-03-23
Add support for iMXRT1061 and 1064

### LINKED ISSUE #105: iMXRT1064 support
We should publish the imxrt-hal with support for the IMXRT1064 device if it is ready.

Continuing discussion from [ imxrt-rs imxrt-boot-gen#7](https://github.com/imxrt-rs/imxrt-boot-gen/issues/7).

> I'm curious as to how to interface that on-board flash. I can't easily find any NXP docs that say "on-board flash starts at this address." (Since the 1064 RM notes that all FlexSPI2 pin muxing registers are "reserved," I'm guessing that on-board flash is available via FlexSPI2.)

The on-board iMXRT1064 flash is at 0x70000000 upto 0x703FFFFF. [This](https://www.nxp.com/docs/en/application-note/AN12290.pdf) document describes the required changes when moving from an iMXRT1060 series device.

The onboard flash does use the FlexSPI2. The onboard flash is actually an embedded [Winbond W25Q32JV](https://www.winbond.com/resource-files/w25q32jv%20spi%20revc%2008302016.pdf)

mciantyre:
The PR adds two new features,

- `"imxrt1061"`
- `"imxrt1064"`

for supporting the 1061 and 1064 MCUs. These should be trivial to support, since they're so similar to the 1062.

Notable differences:

- From brief inspection of the data sheet, the 1061 is missing the LCD/CSI/PXP peripherals. We don't have drivers for these peripherals today, so there's no concern. If they were to be added to the 0.4 HAL, we will have the feature flags to selectively enable the drivers.
- 1064 has 4MB on-board flash connected to FlexSPI2. Given the allocation, the IOMUXC marks the FlexSPI2 registers as "reserved." No problem there, since the `imxrt-iomuxc` crate doesn't have FlexSPI pin APIs. Only other safe API is available from the RAL, which already handles this difference.

Note that this PR targets the 0.4 branch, and may become the 0.4.3 release. Closes #105.


bfrog:
Fantastic, always great to see new people trying this out

PR #107, opened 2021-03-28
Implement ADC DMA source and wrapper

lkolbly:
This commit makes the ADCs a possible source for DMA operations, and creates a
user-friendly wrapper class (StreamingAdc) which can abstract the necessary
buffer wrangling.

The StreamingAdc struct may not be a good fit for this crate (after making it, it seems to be a bit niche - supporting only the interrupts-with-256-element-double-buffers case), if so that's perfectly fine and I can pull it out. I might also be able to make it more better using some subset of:
* Const generics/genericize on buffer type
* Circular buffers
* Gracefully handle overflows
* Allow operation without an interrupt handler

But, it solves _my_ usecase, today at least, and that's the only important one, right? (and I didn't want to spend too much time on it without working toward some specific goal)

Also, this PR is against the `maint-v0.4` branch, because that's what I could get `teensy4-bsp` to work with. If there's a branch of `teensy4-bsp` which works with the `master` branch of `imxrt-hal`, I can try to rebase this.

### DIFF
--- a/imxrt-hal/src/adc.rs
+++ b/imxrt-hal/src/adc.rs
@@ -223,6 +223,64 @@ where
     }
 }
 
+/// Internal trait for determining what DMA request signal to use for
+/// ADC1 or ADC2.
+#[doc(hidden)]
+pub trait AdcDmaSource {
+    /// See table 4-3 of the iMXRT1060 Reference Manual (Rev 2)
+    const SOURCE_REQUEST_SIGNAL: u32;
+}
+
+impl AdcDmaSource for ADC1 {
+    const SOURCE_REQUEST_SIGNAL: u32 = 24;
+}
+
+impl AdcDmaSource for ADC2 {
+    const SOURCE_REQUEST_SIGNAL: u32 = 88;
+}
+
+/// Streaming DMA source for ADCs
+pub struct AdcSource<ADCx, P> {
+    adc: ADC<ADCx>,
+    pin: AnalogInput<ADCx, P>,
+}
+
+impl<ADCx, P> AdcSource<ADCx, P>
+where
+    ADCx: adc::ADC + AdcDmaSource,
+    P: Pin<ADCx>,
+{
+    /// Create a DMA Source object tieing the ADC to the given pin.
+    pub fn new(adc: ADC<ADCx>, pin: AnalogInput<ADCx, P>) -> Self {
+        AdcSource { adc, pin }
+    }
+}
+
+impl<ADCx, P> crate::dma::peripheral::Source<u16> for AdcSource<ADCx, P>
+where
+    ADCx: adc::ADC + AdcDmaSource,
+    P: Pin<ADCx>,
+{
+    type Error = ();
+
+    const SOURCE_REQUEST_SIGNAL: u32 = ADCx::SOURCE_REQUEST_SIGNAL;
+
+    fn source(&self) -> *const u16 {
+        &self.adc.reg.R0 as *const _ as *const u16
+    }
+
+    fn enable_source(&mut self) -> Result<(), Self::Error> {
+        let channel = <P as Pin<ADCx>>::Input::U32;
+        ral::modify_reg!(ral::adc, self.adc.reg, GC, ADCO: 1, DMAEN: 1);
+        ral::modify_reg!(ral::adc, self.adc.reg, HC0, |_| channel);
+        Ok(())
+    }
+
+    fn disable_source(&mut self) {
+        ral::modify_reg!(ral::adc, self.adc.reg, GC, ADCO: 0, DMAEN: 0);
+    }
+}
+
 /// Unclocked ADC modules
 ///
 /// The `Unclocked` struct represents both unconfigured ADC peripherals.
--- a/imxrt-hal/src/dma/peripheral.rs
+++ b/imxrt-hal/src/dma/peripheral.rs
@@ -97,6 +97,9 @@ mod private {
     impl<M> Sealed for uart::Rx<M> where M: crate::iomuxc::consts::Unsigned {}
     impl<M> Sealed for uart::Tx<M> where M: crate::iomuxc::consts::Unsigned {}
 
+    use crate::adc;
+    impl<ADCx, P> Sealed for adc::AdcSource<ADCx, P> {}
+
     use crate::spi;
     impl<M> Sealed for spi::SPI<M> where M: crate::iomuxc::consts::Unsigned {}
 }


lkolbly:
Ah, looks like doctests still compiles `no_run` things, I can fix that if we want to keep that struct.

imxrt-hal/src/adc.rs
```diff
@@ -223,6 +223,205 @@ where
...
+pub struct StreamingAdc<ADCx, P> {
+    buffer0: &'static crate::dma::Buffer<[u16; 256]>,
+    buffer1: &'static crate::dma::Buffer<[u16; 256]>,
+    peripheral: crate::dma::Peripheral<
+        AdcSource<ADCx, P>,
+        u16,
+        crate::dma::Linear<u16>,
+        crate::dma::Linear<u16>,
+    >,
+    unhandled_buffer_count: core::sync::atomic::AtomicUsize,
```
mciantyre, line 346: Since `unhandled_buffer_count` is always modified in `&mut` methods, could this be simplified to a normal `usize`?

imxrt-hal/src/adc.rs
```diff
@@ -223,6 +223,205 @@ where
...
+        let mut rx_buffer = self.peripheral.receive_complete().unwrap();
+        core::mem::swap(&mut self.off_buffer, &mut rx_buffer);
+        self.peripheral.start_receive(rx_buffer).unwrap();
+        if self
+            .unhandled_buffer_count
+            .swap(1, core::sync::atomic::Ordering::SeqCst)
+            > 0
+        {
+            panic!("User is not processing ADC data fast enough!");
+        }
```
mciantyre, line 421: Should we check [`is_receive_complete()`](https://github.com/imxrt-rs/imxrt-hal/blob/8f457edaad37cfd294e6eaeae42605f3214a1b2e/imxrt-hal/src/dma/peripheral.rs#L209-L212) before completing the receive operation, and re-starting the transfer? Otherwise, a `handle_interrupt` call that's invoked before the transfer completes will panic at `receive_complete().unwrap()`.

(Concern is that we could call `handle_interrupt()` prematurely in an ISR. Consider the interrupt in the `StreamingAdc` example, and pretend that we're also using DMA channel 24 for something else. As written, we would invoke `handle_interrupt()` when channel 24 completes, which isn't necessarily when channel 8 completes.)

mciantyre (commented):
Thanks for exploring DMA-powered analog I/O! Happy to see that the DMA abstractions work for another peripheral. `AdcSource` and it associated code looks good to me.

I agree that, in its current form, `StreamingAdc` is a bit specific for the HAL. I also see some patterns that might not work, but I'm looking for your insight. If it's not too much work to generalize, I'm happy to support something like `StreamingAdc` in the HAL. Otherwise, we can merge the foundational work that lets us implement `StreamingAdc` outside of the HAL.

> Also, this PR is against the maint-v0.4 branch, because that's what I could get teensy4-bsp to work with. If there's a branch of teensy4-bsp which works with the master branch of imxrt-hal, I can try to rebase this.

You're targeting the correct branch. Since I'm still supporting the 0.4 HAL, I haven't updated the Teensy 4's BSP to use the HAL's main branch. I'll take care of porting your feature to the HAL's main branch.

imxrt-hal/src/adc.rs
```diff
@@ -223,6 +223,205 @@ where
...
+            off_buffer,
+            _pin: PhantomData,
+        }
+    }
+
+    /// Begin transferring samples from the ADC. Once samples are flowing, the
+    /// user must call poll() frequently enough to keep up with the generated
+    /// samples, or else the interrupt handler will panic.
+    pub fn start(&mut self) {
+        let mut buffer = crate::dma::Linear::new(&self.buffer1).unwrap();
```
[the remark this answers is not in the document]
mciantyre: > I'm afraid that calling start() twice, where buffer1 refers to a previously-used Buffer, will result in a panic. Have we seen this in practice?

If `start()` should only be called once, we could document that, or express the behavior with typestates.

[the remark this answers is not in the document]
lkolbly: Yeah, so far I've only used it letting it run continuously for eternity, so starting hasn't had to happen twice.

I think a `Circular` would be the way to go, I dislike how the double-buffering has to restart the DMA buffer periodically. That's how I was originally doing it, but I was having trouble getting `Circular` to notice that the DMA engine added elements to it. Is there some sort of poll() method I need to call, or should it in principle happen automatically?

A `impl Iterator<Item = u16>` would work fine, it could be the `Drain` iterator, worst case the user just stack-allocates a buffer for whatever they need. The only alternative I can think of would be `Fn(&[u16], &[u16]) -> usize`, returning the number of elements consumed, but that's a bit more complicated (and the second argument would usually be an empty slice anyway).

imxrt-hal/src/adc.rs
lkolbly: My concern is since it gets modified from the interrupt, it doesn't _really_ have an exclusive borrow to the object.

Not sure what the Rust-y way to fix that is. Maybe make both these methods `&self` methods, and use an `UnsafeCell`?

lkolbly: Of course, using `Circular` might circumvent this, since then we shouldn't need an ISR to fiddle with the buffers at all, there's only the user poll()ing as fast as they need.

imxrt-hal/src/adc.rs
lkolbly: > Should we check is_receive_complete() before completing the receive operation, and re-starting the transfer?

Yes! Good catch.

lkolbly:
> I agree that, in its current form, StreamingAdc is a bit specific for the HAL. I also see some patterns that might not work, but I'm
> looking for your insight. If it's not too much work to generalize, I'm happy to support something like StreamingAdc in the HAL.
> Otherwise, we can merge the foundational work that lets us implement StreamingAdc outside of the HAL.

Awesome, thanks! Tell you what, I'll steal your insights on what direction would make sense to make it better, and then I'll update this PR to be just the `AdcSource`, and I'll make a new PR for the `StreamingAdc` once I've cleaned it up a bit. (although, with `Circular`, maybe it's simple enough to justify just being an example in the docs somewhere, most of what it does now is tracking the two buffers)

I may do that last part contemporaneously with SAI support, that's next on my project todo list and it seems like it _may_ deserve something similar.

imxrt-hal/src/adc.rs
mciantyre: For a `StreamingAdc` with mutable methods shared with an ISR, I think we could guarantee exclusive mutable access using a combination of [`Mutex`](https://docs.rs/cortex-m/0.7.2/cortex_m/interrupt/struct.Mutex.html), `RefCell`, and `Option`.

- `Mutex` for exclusion between polling loop and ISR
- `RefCell` for interior mutability required by `Mutex`
- `Option` because it needs setup in `main()` before it's stuffed into a `static`

The types might be gnarly! But there's no unsafe, and it makes sure we have the exclusive access we say we need with `&mut self` methods.

```rust
use core::cell::RefCell;
use cortex_m::interrupt::Mutex;

type MyAdc = Mutex<RefCell<Option<StreamingAdc>>>;
static MY_STREAMING_ADC: MyAdc = Mutex::new(RefCell::new(None));

#[cortex_m_rt::interrupt]
fn DMA8_DMA24() {
    cortex_m::interrupt::free(|cs| {
        // Mutex<RefCell<...>> -> RefCell<...>
        let adc = MY_STREAMING_ADC.borrow(cs);
        // RefCell<Optional<...>> -> &mut Optional<...>
        let mut adc = adc.borrow_mut();
        // &mut Optional<StreamingAdc> -> &mut StreamingAdc
        let mut adc = adc.as_mut().unwrap();
        // Swap buffers, etc...
    });
}

fn main() {
    // Setup ADC n'at...

    // Prepare shared state between ISR, polling loop...
    cortex_m::interrupt::free(|cs| {
        let adc = MY_STREAMING_ADC.borrow(cs);
        let mut adc = adc.borrow_mut();
        *adc = Some(the_adc);
    });

    // Polling loop...
    cortex_m::interrupt::free(|cs| {
        // As before...
    });
}
```

imxrt-hal/src/adc.rs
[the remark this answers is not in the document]
mciantyre: >  I was having trouble getting Circular to notice that the DMA engine added elements to it.

Sorry for the trouble! We're calling [`reserve()`](https://docs.rs/imxrt-hal/0.4.2/imxrt_hal/dma/struct.Circular.html#method.reserve) on the circular buffer before submitting it to the channel? That should signal how much data we want to receive in the next DMA transfer. After that, it should be automatic. See the [`memcpy` example](https://github.com/mciantyre/teensy4-rs/blob/73433b3ae6cbd7754729d0dcecca61bbfe0419a2/examples/bsp/src/bin/dma_memcpy.rs#L73) in the Teensy 4 BSP for a demo.

mciantyre:
Sounds great! Thanks again for working though this. Let us know if we can help with any of the ADC, DMA, or SAI work.

PR #108, opened 2021-04-23
Fix PWM driver

mciantyre:
This PR fixes a regression in the PWM driver, introduced in 0.4.0.

In the 0.3 HAL, we relied on the user to set the pin alternate
settings. But after introducing the IOMUXC crate in 0.4, we made pin
muxing part of the driver's responsibility, and added calls to prepare
pins. We missed the requirement in the PWM driver. This commit
corrects the PWM driver, which has been broken since the 0.4 release.

Tested in the teensy4-rs repo. Users who are not able to adopt the
next patch release could add equivalent iomuxc calls before
constructing their PWM driver.

### DIFF
--- a/imxrt-hal/CHANGELOG.md
+++ b/imxrt-hal/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


bfrog (approved):
LGTM

PR #110, opened 2021-06-30
GPIO functions to enable / configure interrupts

### LINKED ISSUE #109: Support GPIO input interrupts
The 0.4 HAL does not have any way to support interrupts on GPIO inputs. This also applies to any previous version of `imxrt-hal`. Add and document an API that supports

- enabling / disabling interrupts on a GPIO input
- selecting the various edge / level configurations

Since this should be a brand-new feature, it should be possible to release in 0.4.5 without issues.

algesten:
cortex_m_rt provides the hook for the ISR and the `NVIC::unmask`,
however to configure a GPIO interrupt there are some additional
steps needed.

This commit provides set_interrupt_enable, set_interrupt_configuration
and clear_interrupt_status (and some additional query functions).

Close #109

### DIFF
--- a/imxrt1060-hal/src/gpio.rs
+++ b/imxrt1060-hal/src/gpio.rs
@@ -71,6 +71,13 @@ where
         1u32 << <P as Pin>::Offset::USIZE
     }
 
+    /// Returns the ICR field offset for this GPIO
+    ///
+    /// ICR is "Interrupt Clear Register"
+    fn icr_offset(&self) -> usize {
+        (<P as Pin>::Offset::USIZE % 16) * 2
+    }
+
     /// The return is a non-zero number, since the GPIO identifiers
     /// start with '1.'
     #[inline(always)]
@@ -209,6 +216,84 @@ where
     pub fn is_set(&self) -> bool {
         // Safety: read is atomic
         unsafe { ral::read_reg!(ral::gpio, self.register_block(), PSR) & self.mask() != 0 }
+    }
+}
+
+/// GPIO input interrupt configurations.
+///
+/// These configurations do not take effect until
+/// GPIO input interrupts are enabled.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u32)]
+pub enum InterruptConfiguration {
+    // TODO isn't it a bit repetitive to have "Sensitive" on all here?
+    LowLevelSensitive = 0,
+    HighLevelSensitive = 1,
+    RisingEdgeSensitive = 2,
+    FallingEdgeSensitive = 3,
+    EitherEdgeSensitive = 4,
+}
+
+impl<P> GPIO<P, Input>
+where
+    P: Pin,
+{
+    /// Enable (`true`) or disable (`false`) interrupts for this GPIO input.
+    pub fn set_interrupt_enable(&mut self, enable: bool) {
+        cortex_m::interrupt::free(|_| unsafe {
+            ral::modify_reg!(ral::gpio, self.register_block(), IMR, |imr| if enable {
+                imr | self.mask()
+            } else {
+                imr & !self.mask()
+            })
+        });
+    }
+
+    /// Indicates if interrupts are (`true`) or are not (`false`) enabled for this GPIO input.
+    pub fn is_interrupt_enabled(&self) -> bool {
+        unsafe { ral::read_reg!(ral::gpio, self.register_block(), IMR) & self.mask() != 0u32 }
+    }
+
+    /// Set the interrupt configuration for this GPIO input.
+    pub fn set_interrupt_configuration(&mut self, interrupt_configuration: InterruptConfiguration) {
+        if InterruptConfiguration::EitherEdgeSensitive == interrupt_configuration {
+            unsafe {
+                ral::modify_reg!(ral::gpio, self.register_block(), EDGE_SEL, |edge_sel| {
+                    edge_sel | self.mask()
+                });
+            }
+        } else {
+            unsafe {
+                ral::modify_reg!(ral::gpio, self.register_block(), EDGE_SEL, |edge_sel| {
+                    edge_sel & !self.mask()
+                });
+            }
+            // TODO is it ok to assume the numeric value of the enum is the icr?
+            let icr = interrupt_configuration as u32;
+            let icr_offset = self.icr_offset();
+            let icr_modify = |reg| reg & !(0b11 << icr_offset) | (icr << icr_offset);
+            if <P as Pin>::Offset::USIZE < 16 {
+                unsafe {
+                    ral::modify_reg!(ral::gpio, self.register_block(), ICR1, icr_modify);
+                }
+            } else {
+                unsafe {
+                    ral::modify_reg!(ral::gpio, self.register_block(), ICR2, icr_modify);
+                }
+            }
+        }
+    }
+
+    /// Indicates whether this GPIO input triggered an interrupt.
+    pub fn is_interrupt_status(&self) -> bool {
+        unsafe { ral::read_reg!(ral::gpio, self.register_block(), ISR) & self.mask() != 032 }
+    }
+
+    /// Clear the interrupt status flag.
+    pub fn clear_interrupt_status(&mut self) {
+        // TODO is it okay to unconditonally write to clear, or do we need to
+        // check it's actually set first?
+        unsafe { ral::write_reg!(ral::gpio, self.register_block(), ISR, self.mask()) }
     }
 }
 


imxrt1060-hal/src/gpio.rs
```diff
@@ -212,6 +219,84 @@ where
...
 }
 
+/// GPIO input interrupt configurations.
+///
+/// These configurations do not take effect until
+/// GPIO input interrupts are enabled.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u32)]
+pub enum InterruptConfiguration {
+    // TODO isn't it a bit repetitive to have "Sensitive" on all here?
```
algesten, line 229: Style question here. Maybe we drop `Sensitive`?

imxrt1060-hal/src/gpio.rs
```diff
@@ -212,6 +219,84 @@ where
...
+                    edge_sel | self.mask()
+                });
+            }
+        } else {
+            unsafe {
+                ral::modify_reg!(ral::gpio, self.register_block(), EDGE_SEL, |edge_sel| {
+                    edge_sel & !self.mask()
+                });
+            }
+            // TODO is it ok to assume the numeric value of the enum is the icr?
```
algesten, line 271: In your async code you had an explicit mapping from the enum to the `icr`. Should we keep that?

imxrt1060-hal/src/gpio.rs
```diff
@@ -212,6 +219,84 @@ where
...
+    }
+
+    /// Indicates whether this GPIO input triggered an interrupt.
+    pub fn is_interrupt_status(&self) -> bool {
+        unsafe { ral::read_reg!(ral::gpio, self.register_block(), ISR) & self.mask() != 032 }
+    }
+
+    /// Clear the interrupt status flag.
+    pub fn clear_interrupt_status(&mut self) {
+        // TODO is it okay to unconditonally write to clear, or do we need to
```
algesten, line 294: Should we check whether the bit really is set first? Is it ok to call this when the bit is not clear?

imxrt1060-hal/src/gpio.rs
```diff
@@ -71,6 +71,13 @@ where
         1u32 << <P as Pin>::Offset::USIZE
     }
 
+    /// Returns the ICR field offset for this GPIO
+    ///
+    /// ICR is "Interrupt Clear Register"
```
mciantyre, line 76: ICR is the interrupt _configuration_ register, accepting `InterruptConfiguration` enum values. We'll clear interrupts with the status register, ISR.
```suggestion
    /// ICR is "Interrupt Configuration Register"
```

imxrt1060-hal/src/gpio.rs
mciantyre: Yea it's repetitive. OK with me to drop it!

I copied it straight from the reference manual, but we should't need it to understand these values.

imxrt1060-hal/src/gpio.rs
mciantyre: I like your `as u32` approach! Wasn't relying on a `#[repr(u32)]` enum in the async code, though an `as` cast might have yielded the same values.

imxrt1060-hal/src/gpio.rs
mciantyre: The ISR register is "write-1-clear" (W1C). Once this write completes, the bit is cleared in the register.

I think we can skip the check. The user can check if it's set using `is_interrupt_status()`. If the bit is 0, writing a 1 to clear the bit has no effect.

imxrt1060-hal/src/gpio.rs
```diff
@@ -212,6 +219,84 @@ where
...
+            unsafe {
+                ral::modify_reg!(ral::gpio, self.register_block(), EDGE_SEL, |edge_sel| {
+                    edge_sel | self.mask()
+                });
+            }
+        } else {
+            unsafe {
+                ral::modify_reg!(ral::gpio, self.register_block(), EDGE_SEL, |edge_sel| {
+                    edge_sel & !self.mask()
+                });
```
mciantyre, line 269: We'll need critical sections around these `modify_reg!` macros. It's due to a design limitation in today's GPIO driver. Since the `EDGE_SEL` register is shared across all GPIO[N] pins, there's a chance for a RMW to be preempted, which could result in an inconsistent state.

The critical sections aren't the best, nor the perfect defense, but I think we need a breaking change and driver re-design to correct it.

imxrt1060-hal/src/gpio.rs
```diff
@@ -212,6 +219,84 @@ where
...
+            let icr = interrupt_configuration as u32;
+            let icr_offset = self.icr_offset();
+            let icr_modify = |reg| reg & !(0b11 << icr_offset) | (icr << icr_offset);
+            if <P as Pin>::Offset::USIZE < 16 {
+                unsafe {
+                    ral::modify_reg!(ral::gpio, self.register_block(), ICR1, icr_modify);
+                }
+            } else {
+                unsafe {
+                    ral::modify_reg!(ral::gpio, self.register_block(), ICR2, icr_modify);
```
mciantyre, line 281: We'll need critical sections here, too. See previous comment.

mciantyre (commented):
Thanks again for taking this on! GPIO input methods look OK.

I think we'll need to update the `copy_settings` method to handle the

- ICR
- EDGE_SEL
- IMR

registers. We want to make sure that, if we change from a normal GPIO2_x to a fast GPIO7_x, the settings from GPIO2_x are maintained in GPIO7_x.

We also might want to disable interrupts if we transition from input => output, though this might already be prevented by the hardware (will need to check reference manual).

imxrt1060-hal/src/gpio.rs
```diff
@@ -212,6 +219,84 @@ where
...
+            })
+        });
+    }
+
+    /// Indicates if interrupts are (`true`) or are not (`false`) enabled for this GPIO input.
+    pub fn is_interrupt_enabled(&self) -> bool {
+        unsafe { ral::read_reg!(ral::gpio, self.register_block(), IMR) & self.mask() != 0u32 }
+    }
+
+    /// Set the interrupt configuration for this GPIO input.
```
algesten, line 257: If the user does two calls like this:

```rust
pin.set_interrupt_configuration(InterruptConfiguration::RisingEdge);

…

pin.set_interrupt_configuration(InterruptConfiguration::EitherEdge);
```

That means we have modified ICR1 or ICR2. Do we need to undo that modification to go back to `EitherEdge`?

imxrt1060-hal/src/gpio.rs
```diff
@@ -160,12 +167,26 @@ where
         // The input / output direction. When switching across fast / normal, keep the
         // same direction.
         copy_bits!(GDIR, self.mask());
+
+        // Interrupt configuration is preserved when switching fast / normal.
+        if <P as Pin>::Offset::USIZE < 16 {
+            copy_bits!(ICR1, self.mask());
+        } else {
+            copy_bits!(ICR2, self.mask());
+        }
```
algesten, line 176: Is this necessary?!

GPIO1-5 are the ones that can transition to fast (GPIO6-9). Are they maybe all in `ICR1` so we can skip the offset check?

mciantyre: Selecting either ICR1 or ICR2, based on the pin number, is correct. The ICR field for `GPIO1_23` is in ICR2, at bit 15. If we become a fast GPIO (`GPIO6_23`), the ICR field is still in ICR2, at bit 15. It's just in a different register block.

imxrt1060-hal/src/gpio.rs
```diff
@@ -160,12 +167,26 @@ where
         // The input / output direction. When switching across fast / normal, keep the
         // same direction.
         copy_bits!(GDIR, self.mask());
+
+        // Interrupt configuration is preserved when switching fast / normal.
+        if <P as Pin>::Offset::USIZE < 16 {
+            copy_bits!(ICR1, self.mask());
+        } else {
+            copy_bits!(ICR2, self.mask());
```
mciantyre, line 175: We'll need to use ICR masks here. ICR fields are two bits wide. To hold all the interrupt configurations for up to 32 GPIOs, we need 64 bits. That's why there's two 32-bit ICR registers.

```suggestion
            copy_bits!(ICR1, 0b11 << self.icr_offset());
        } else {
            copy_bits!(ICR2, 0b11 << self.icr_offset());
```

imxrt1060-hal/src/gpio.rs
mciantyre: No need to undo anything. When `EitherEdge` is activated, the hardware ignores whatever is in the ICR register. Here's the snippet from the reference manual.

image

mciantyre (commented):
Looking good! The only thing we need to check is the ICR mask supplied to `copy_bits!`.

imxrt1060-hal/src/gpio.rs
algesten: Very good!

algesten:
Cool! To be consistent should we rename icr_offset to icr_mask?

mciantyre:
> Cool! To be consistent should we rename icr_offset to icr_mask?

Great call, good with me!

mciantyre:
Thanks again for working through this. Latest commits add an ICR mask method, and add documentation & examples. Are there any other TODOs for this PR?

mciantyre/teensy4-rs#104 proposes an RTIC-based example that seems to work on my setup. If you have other examples that you'd like to contribute, I'm happy to support them.

mciantyre:
Would also appreciate any feedback on docs in 56fb24b.

PR #113, opened 2021-12-02
Follow rand_core API

AlyoshaVasilieva:
`rand_core` [requires that custom error codes start at `0xC000_0000`](https://docs.rs/rand_core/0.6.3/rand_core/struct.Error.html#associatedconstant.CUSTOM_START), which I missed when I wrote the impl


mciantyre:
Great catch. Thanks for the fix! I'll release in 0.4.5.

PR #115, opened 2022-01-04
Add tempmon abstruction

Alexander89:
I got a little bit side trapped by the *Temperature Monitor (TEMPMON)*

I thought it is barely simple and a great start in getting my hands dirty with the `imxrt-ral` ecosystem.

The interrupts are not really handy.

---

# Open to discuss

I played around with the UX a little bit and finally, it is OK. 

In my test app, I implemented a static alarm flag to note, when I have to enable the interrupts again.
Another method would be, to stop the measurement process, but as soon you read the temperature, the interrupt is triggered again. 

```rust
// skip rest

let mut t = peripherals.tempmon.init_with_measure_freq(0x100);
t.set_alarm_values(40.0, 70.0, 85.0);
let (low_alarm, high_alarm, panic_alarm) = t.alarm_values();
log::debug!("low {}, high {}, panic {}", low_alarm, high_alarm, panic_alarm);

// v2: AtomicBool 
static mut ALARM: bool = false;
#[cortex_m_rt::interrupt]
fn TEMP_LOW_HIGH() {
  unsafe { ALARM = true; }
  TempMon::disable_interrupts(true, false);
  cortex_m::asm::dsb();
}

loop {
  let temp = t.get_temp();
  if temp > low_alarm && temp < high_alarm && unsafe { alarm } == true {
    unsafe { alarm = false; }
    t.enable_interrupts(true, true);
  }
}
```

If anybody has a great idea on how to improve it, please let me know :smiley:

### DIFF
--- a/imxrt-hal/src/tempmon.rs
+++ b/imxrt-hal/src/tempmon.rs
@@ -2,16 +2,19 @@
 
 //! Temperature Monitor (TEMPMON)
 //!
-//! # Example
+//! IMPORTANT NOTE: The temperature sensor uses and assumes that the bandgap
+//! reference, 480MHz PLL and 32KHz RTC modules are properly programmed and fully
+//! settled for correct operation.
 //!
-//! Manually trigger read
+//! # Example 1
+//!
+//! Manually triggered read
 //!
 //! ```no_run
 //! use imxrt1060_hal;
 //! use embedded_hal::timer::CountDown;
 //!
 //! let mut peripherals = imxrt1060_hal::Peripherals::take().unwrap();
-//! usb_io::init();
 //!
 //! let (_, ipg_hz) = peripherals.ccm.pll1.set_arm_clock(
 //!     imxrt1060_hal::ccm::PLL1::ARM_HZ,
@@ -25,26 +28,83 @@
 //!     imxrt1060_hal::ccm::perclk::CLKSEL::IPG(ipg_hz),
 //! );
 //!
-//! log::debug!("init temperature monitor");
-//! let mut mon = peripherals.tempmon.init();
-//! t.set_alarm_values()
-//! UnInitialized::new(peripherals.temp)
-//! todo!()
+//! // init temperature monitor
+//! let mut temp_mon = peripherals.tempmon.init();
+//! loop {
+//!     let temperature = temp_mon.measure_temp();
+//! }
+//! ```
+//!
+//! # Example 2
+//!
+//! Non-blocking reading
+//!
+//! ```no_run
+//! // [Example 1]
+//!
+//! // Init temperature monitor with 8Hz measure freq
+//! // 0xffff = 2 Sec. Read more at `measure_freq()`
+//! let mut temp_mon = peripherals.tempmon.init_with_measure_freq(0x1000);
+//! temp_mon.start();
+//!
+//! let last_temp = 0.0_f32
+//! loop {
+//!     // Get the last temperature read by the measure_freq
+//!     let temp = t.get_temp();
+//!     if last_temp != temp {
+//!         // temperature changed
+//!         last_temp = temp;
+//!     }
+//!     // do something else
+//! }
+//! ```
+//!
+//! # Example 3
+//!
+//! Low and high temperature Interrupt
+//!
+//! ```no_run
+//! // [Example 1]
+//!
+//! // init temperature monitor with 8Hz measure freq
+//! // 0xffff = 2 Sec. Read more at `measure_freq()`
+//! let mut temp_mon = peripherals.tempmon.init_with_measure_freq(0x1000);
+//!
+//! // set low_alarm, high_alarm, and panic_alarm
+//! temp_mon.set_alarm_values(-5.0, 65.0, 95.0);
+//!
+//! // use values from registers if you like to compare it somewhere
+//! let (low_alarm, high_alarm, panic_alarm) = t.alarm_values();
+//!
+//! // enables interrupts for low_high_alarm and not for panic_alarm
+//! temp_mon.enable_interrupts(true, false);
+//! temp_mon.start();
+//!
+//! #[cortex_m_rt::interrupt]
+//! fn TEMP_LOW_HIGH() {
+//!     // disable the interrupt to avoid endless triggers
+//!     TempMon::disable_interrupts(true, false);
+//!
+//!     // don't forget to enable it after the temperature is back to normal
+//! }
 //! ```
 
 use crate::ral;
 
-/// An unclocked periodic interrupt timer module
+/// An Uninitialized temperature monitor module
 ///
-/// In order to activate the PIT, we must pass in the
-/// configured object returned from the CCM's perclk module.
-pub struct UnInitialized(ral::tempmon::Instance);
+/// The temperature sensor uses and assumes that the bandgap
+/// reference, 480MHz PLL and 32KHz RTC modules are properly
+/// programmed and fully settled for correct operation.
+pub struct Uninitialized(ral::tempmon::Instance);
 
-impl UnInitialized {
+impl Uninitialized {
+    /// assign the tempmon Instance to this temperature monitor wrapper.
     pub fn new(base: ral::tempmon::Instance) -> Self {
         Self(base)
     }
 
+    /// Initialize the temperature monitor.
     pub fn init(self) -> TempMon {
         let calibration = unsafe { ral::read_reg!(ral::ocotp, OCOTP, ANA1) };
 
@@ -84,27 +144,48 @@ impl UnInitialized {
 }
 
 /// A Temperature Monitor (TEMPMON)
+///
+/// # Example 1
+///
+/// ```no_run
+/// use imxrt1060_hal;
+/// use embedded_hal::timer::CountDown;
+///
+/// let mut peripherals = imxrt1060_hal::Peripherals::take().unwrap();
+/// let (_, ipg_hz) = peripherals.ccm.pll1.set_arm_clock(
+///     imxrt1060_hal::ccm::PLL1::ARM_HZ,
+///     &mut peripherals.ccm.handle,
+///     &mut peripherals.dcdc,
+/// );
+/// let mut cfg = peripherals.ccm.perclk.configure(
+///     &mut peripherals.ccm.handle,
+///     imxrt1060_hal::ccm::perclk::PODF::DIVIDE_3,
+///     imxrt1060_hal::ccm::perclk::CLKSEL::IPG(ipg_hz),
+/// );
+///
+/// // init temperature monitor
+/// // consider using init_with_measure_freq
+/// let mut temp_mon = peripherals.tempmon.init();
+/// loop {
+///     let temperature = temp_mon.measure_temp();
+/// }
+/// ```
 pub struct TempMon {
     base: ral::tempmon::Instance,
-    pub scaler: f32,
-    pub hot_count: u16,
-    pub hot_temp: f32,
+    scaler: f32,
+    hot_count: u16,
+    hot_temp: f32,
 }
 
 impl TempMon {
-    pub fn convert(&self, temp_cnt: u16) -> f32 {
+    /// converts the temp_cnt into a human readable temperature [C°]
+    fn convert(&self, temp_cnt: u16) -> f32 {
         let n_meas: f32 = (temp_cnt - self.hot_count).into();
         self.hot_temp - n_meas * self.scaler
     }
 
-    #[allow(dead_code)]
+    /// decode the temp_value into measurable bytes (f32 -> u32)
     fn decode(&self, temp_value: f32) -> u32 {
-        // let n_meas: f32 = (temp_cnt - self.hot_count).into();
-        // let value = self.hot_temp - n_meas * self.scaler;
-
-        // let temp_value - self.hot_temp =  0 - (temp_cnt - self.hot_count) * self.scaler;
-        // let ((temp_value - self.hot_temp) / self.scaler) = -temp_cnt + self.hot_count);
-        // let temp_cnt + ((temp_value - self.hot_temp) / self.scaler) = self.hot_count);
         let v: i32 = ((temp_value - self.hot_temp) / self.scaler) as i32;
         (self.hot_count as i32 - v) as u32
     }
@@ -139,7 +220,6 @@ impl TempMon {
         );
     }
 
-    #[allow(dead_code)]
     /// Stops the measurement process. This only has an effect If the measurement
     /// frequency is not zero.
     pub fn stop(&mut self) {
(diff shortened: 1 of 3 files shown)


teburd (commented):
Thank you for contributing!

I think in general the idea is really nice, and the API seems reasonable. A few things though..

The usage of floats in the API is, in my opinion, less than ideal. It's ok if you have a few helper functions alongside this as const fn's if you like, but float as a whole tend to be troublesome. Yes this micro has a fpu so perhaps its less of an issue, but generally speaking, on a microcontroller and it's SDK you aren't likely going to see floats in use.

Does it make sense as well to make the powered state of this peripheral part of its type state to help the programmer ensure the peripheral is in the desired state at various code points where it might be used?

Consider someone calls get_temp(), or measure_temp on an unpowered tempmon, what happens if its off?

imxrt-hal/src/tempmon.rs
```diff
@@ -0,0 +1,332 @@
...
+    pub fn init(self) -> TempMon {
+        let calibration = unsafe { ral::read_reg!(ral::ocotp, OCOTP, ANA1) };
+
+        let n1_room_count = (calibration >> 20) as u16;
+        let t1_room_temp = 25_u16;
+        let n2_hot_count = ((calibration >> 8) & 0xFFF) as u16;
+        let t2_hot_temp = (calibration & 0xFF) as u16;
+
+        // Tmeas = T2 - (Nmeas - N2) * ((T2 – T1) / (N1 – N2))
+        let t_dif: f32 = (t2_hot_temp - t1_room_temp).into();
```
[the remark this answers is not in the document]
Alexander89: good idea. I used something like this a while ago for a DS18B20. 

I'll change this :heart:

Alexander89:
> Consider someone calls get_temp(), or measure_temp on an unpowered tempmon, what happens if it's off?

Actually, `get_temp()` still works or it just returns old data. So, somehow it doesn't really work, though.
But measure_temp() will just loop to the end of time.

I will add a *power_on* check and as you mention and I turn the values into a Result<i32>.

### One additional question:
Should I change the messure_temp() into non-blocking and return `WouldBlock` in the blocking case. (`nb::block!()` compatible)

teburd:
> Should I change the messure_temp() into non-blocking and return `WouldBlock` in the blocking case. (`nb::block!()` compatible)

Seems reasonable, a Result return here makes sense as it is fallible if the device isn't on. Perhaps using nb here would be nice, I haven't used it in an async way myself though so I can't comment one way or the other here

imxrt-hal/src/tempmon.rs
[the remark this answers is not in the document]
Alexander89: fixed with °mC

Works fine.

Alexander89:
ich changed the math to °mC and adopted the docs to that.

In addition, I implemented an error handling to deal with the power_down state.

The temp_mon.measure_temp() is now compatible to `nb::block!()`

I also ran `cargo clippy` and `cargo fmt`, I hope the pipeline is now happy :-)

Alexander89:
@mciantyre, I created an example app to test this code, should I add it to [teensy4-rs/examples](https://github.com/mciantyre/teensy4-rs/tree/master/examples)?

imxrt-hal/src/tempmon.rs
```diff
@@ -0,0 +1,395 @@
...
+//!         // temperature in mC (1°C = 1000°mC)
+//!     }
+//! }
+//! ```
+//!
+//! ## Example 2
+//!
+//! Non-blocking reading
+//!
+//! ```ignore
```
mciantyre, line 44: Could we make this a `no_run` example, and include the necessary code to make it compile? We could [hide portions of the example 1 code](https://doc.rust-lang.org/rustdoc/documentation-tests.html#hiding-portions-of-the-example) to keep the example focused on the new content.

Question also applies to example 3, below.

imxrt-hal/src/tempmon.rs
```diff
@@ -0,0 +1,395 @@
...
+    /// Starts the measurement process. If the measurement frequency is zero, this
+    /// results in a single conversion.
+    pub fn start(&mut self) -> nb::Result<(), TempMonError> {
+        if self.is_powered_up() {
+            ral::write_reg!(
+                ral::tempmon,
+                self.base,
+                TEMPSENSE0,
+                MEASURE_TEMP: 1
+            );
```
mciantyre, line 271: This `write_reg!` sets the `MEASURE_TEMP` bit high, _and sets all other register bits to zero_. We're similarly using `write_reg!` throughout this implementation to change one field in a register. Is this causing any issues in practice?

Would `modify_reg`, which preserves the unspecified fields, be more appropriate? Or, could we use the various `*_SET`, `*_CLR`, and `*_TOGGLE` registers with `write_reg!` to achieve our goals? It might be better to write this 1 to `TEMPSENSE0_SET`, since the hardware will preserve the other bits. (I'd recommend using the set/clear/toggle registers when possible. We could then discuss `&mut self` vs `&self` receivers for these methods.)

imxrt-hal/src/tempmon.rs
```diff
@@ -0,0 +1,395 @@
...
+    /// decode the temp_value into measurable bytes
+    ///
+    /// param **temp_value_mc**: in °mC (1/1000°C)
+    ///
+    fn decode(&self, temp_value_mc: i32) -> u32 {
+        let v = (temp_value_mc - self.hot_temp) / self.scaler;
+        (self.hot_count - v) as u32
+    }
+
+    /// triggers a new measurement and waits until it's finished
```
mciantyre, line 217: Are we still waiting with the adoption of `WouldBlock`?
```suggestion
    /// triggers a new measurement
```

imxrt-hal/src/tempmon.rs
```diff
@@ -0,0 +1,395 @@
...
+                self.measurement_active = true;
+            }
+
+            // if the measurement is not finished or not started
+            // i.MX Docs: This bit should be cleared by the sensor after the start of each measurement
+            if ral::read_reg!(ral::tempmon, self.base, TEMPSENSE0, FINISHED == 0) {
+                // this could be triggered again without any effect
+                Err(nb::Error::WouldBlock)
+            } else {
+                self.measurement_active = false;
```
mciantyre, line 241: Is it important to maintain this state in RAM? This seems like information we already have in the `MEASURE_TEMP` field. Instead of setting this bool to `false`, we could set `MEASURE_TEMP = 0`.

mciantyre (commented):
Thanks for the new peripheral. A few questions and observations from my first review.

> I created an example app to test this code, should I add it to teensy4-rs/examples?

Happy to maintain it there. Thanks!

imxrt-hal/src/tempmon.rs
Alexander89: I changed it, please let me know if this is how it should be. I didn't find an example.

I think it should be `&mut self.` 
Otherwise, somebody could change the behavior of the tempmon without having a mutable reference.

Alexander89: > I changed it, please let me know if this is how it should be. I didn't find an example

Found it, `modify_reg!()` is what I was looking for

teburd (approved):
I believe all the vital changes have been made, some wording could be improved in some comments such as describing parameters and returns, especially units involved when they aren't clear.

imxrt-hal/src/tempmon.rs
```diff
@@ -0,0 +1,400 @@
...
+//! // 0xffff = 2 Sec. Read more at `measure_freq()`
+//! let mut temp_mon = peripherals.tempmon.init_with_measure_freq(0x1000);
+//!
+//! // set low_alarm, high_alarm, and panic_alarm
+//! temp_mon.set_alarm_values(-5_000, 65_000, 95_000);
+//!
+//! // use values from registers if you like to compare it somewhere
+//! let (low_alarm, high_alarm, panic_alarm) = temp_mon.alarm_values();
+//!
+//! // enables interrupts for low_high_alarm and not for panic_alarm
```
mciantyre, line 114: > [...] and _not_ for panic_alarm

The `unmask` below appears enable `TEMP_PANIC` interrupts. Is the "not" correct?

```suggestion
//! // enables interrupts for low_high_alarm and panic_alarm
```

imxrt-hal/src/tempmon.rs
```diff
@@ -0,0 +1,400 @@
...
+//!     temp_mon.start();
+//! }
+//!
+//! #[cortex_m_rt::interrupt]
+//! fn TEMP_LOW_HIGH() {
+//!     // disable the interrupt to avoid endless triggers
+//!     unsafe {
+//!         cortex_m::peripheral::NVIC::mask(interrupt::TEMP_LOW_HIGH);
+//!         cortex_m::peripheral::NVIC::mask(interrupt::TEMP_PANIC);
+//!     }
```
mciantyre, line 132: > disable the interrupt to avoid endless triggers

Would disabling tempmon measurements, instead of masking interrupts, be another way to (temporarily) disable the interrupts? If that works, should it be the recommended approach (or at least documented as a different approach)?

(I'm thinking about RTIC users with this suggestion. RTIC users may have relinquished interrupt control to the scheduler. They might appreciate docs that remind them disabling measurements will also disable interrupts.)

---

If keeping the `mask` recommendations, know that [`mask` is safe](https://docs.rs/cortex-m/0.7.4/cortex_m/peripheral/struct.NVIC.html#method.mask), so we can drop the `unsafe` block.

```suggestion
//!     // disable the interrupt to avoid endless triggers
//!     cortex_m::peripheral::NVIC::mask(interrupt::TEMP_LOW_HIGH);
//!     cortex_m::peripheral::NVIC::mask(interrupt::TEMP_PANIC);
```

imxrt-hal/src/tempmon.rs
```diff
@@ -0,0 +1,400 @@
+#![cfg(any(feature = "imxrt1061", feature = "imxrt1062", feature = "imxrt1064"))]
```
mciantyre, line 1: If we're guarding the module inclusion in `lib.rs` with features

```rust
#[cfg(any(feature = "imxrt1061", feature = "imxrt1062", feature = "imxrt1064"))]
 pub mod tempmon;
```

do the guards on this module become redundant?

---

We're not officially supporting any other i.MX RT families besides the 1060 and 1064 parts. However, a quick search through the 1010, 1015, 1020, and 1050 reference manuals indicates that this peripheral exists, and seems similar. (I also did a quick SVD / RAL comparison, and looks like all registers and fields are identical). Are these feature flags necessary? It seems that a contributor who wants to support the 1010 should be able to use this peripheral without issue.

It's OK with me to keep, since this reflects how it was developed and tested. But let us know if you're aware of any limitations; it'll help us later.

imxrt-hal/src/tempmon.rs
```diff
@@ -0,0 +1,400 @@
...
+//! ```
+
+use crate::ral;
+
+/// Indicates that the temperature monitor is powered down.
+///
+/// If you receive this error, `power_up()` the temperature monitor first,
+/// and try again.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct PowerDownError();
```
mciantyre, line 145: An empty tuple allows anyone outside of this module to construct this error. A private field

```rust
pub struct PowerDownError(());
```

prevents users beyond this module creating this error type. Could we change this?

[With your permission](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), I'm happy to make this change myself. Otherwise, this is a note to self; I'll address before merging.

imxrt-hal/src/tempmon.rs
Alexander89: believe it or not, I double-checked it and I didn't know why I assumed this behavior. 

Probably I expected this that way because I start the next measurement with a 1. 
I learned a lot about memory-mapping and how it works. Thanks for your patience :-)

imxrt-hal/src/tempmon.rs
Alexander89: I just changed it, because I didn't really understand how it works. but using `*_SET` and `*_CLR` feels much faster!

imxrt-hal/src/tempmon.rs
Alexander89: I removed it from the module. I had a bad vscode setup so rust-analyzer was confused without it, but not it works even there.

imxrt-hal/src/tempmon.rs
Alexander89: changed. 

Sometimes I like the types in a match statement, but in this case, `Error::Other(_)` is totally fine.

Alexander89:
Thanks for all your patience! I learned a lot about the RAL layer and many other things.
:pray: :rocket: 

I think this is it now :tada:

imxrt-hal/src/tempmon.rs
mciantyre: No problem! Thanks for double checking, and for updating the approach. It wouldn't have surprised me if the behavior was undocumented / missing in the reference manual 😛.

mciantyre (approved):
Looks awesome. Thank you for working through this!

OK with me to merge. Leaving this open in case @teburd wants [any other documentation updates](https://github.com/imxrt-rs/imxrt-hal/pull/115#pullrequestreview-846201604). Otherwise, I'll merge and forward port to the main branch sometime next / the following weekend.

We can use #116 to discuss how we want to release this feature.

imxrt-hal/src/tempmon.rs
```diff
@@ -0,0 +1,390 @@
...
+    fn convert(&self, temp_cnt: i32) -> i32 {
+        let n_meas = temp_cnt - self.hot_count;
+        self.hot_temp - n_meas * self.scaler
+    }
+
+    /// decode the temp_value into measurable bytes
+    ///
+    /// param **temp_value_mc**: in °mC (1/1000°C)
+    ///
+    fn decode(&self, temp_value_mc: i32) -> u32 {
```
teburd, line 239: Can you describe the unit returned

imxrt-hal/src/tempmon.rs
```diff
@@ -0,0 +1,390 @@
...
+    /// scaler * 1000
+    scaler: i32,
+    /// hot_count * 1000
+    hot_count: i32,
+    /// hot_temp * 1000
+    hot_temp: i32,
+}
+
+impl TempMon {
+    /// converts the temp_cnt into a human readable temperature [°mC] (1/1000 °C)
```
teburd, line 229: Can you comment on the units of the parameter and return here as well

teburd (approved):

teburd:
Noted the places where I saw units for parameters and returns might be useful, I believe its mC -> C and vice versa, but would be nice to have it noted there.

PR #121, opened 2022-09-26
RFC: A 0.5 HAL to support more i.MX RT chips

mciantyre:
I've infrequently played around with new `imxrt-hal` ideas. This PR summarizes those ideas. It's a re-write of the 0.4 HAL, a package which only supports 1060 chips. This PR is big and incomplete, but I wanted to talk about ideas with working code.

Unlike today's HAL, this exploratory HAL

- provides common peripherals for _all_ i.MX RT chips supported by `imxrt-ral`, including multi-core 1176 chips.
- maintains hardware examples and tests in tree, and supports development on three different boards.
- provides a path towards a split HAL (#56).

This new HAL supports almost all peripherals from the 0.4 release, but everything is a breaking change.

Asks
----

If you have the time, here's what I'm hoping to get out of this discussion:

- Try out examples on hardware. Better yet, extend the `board` package, and see if you can turn on your board's LED. Ping me below or on the imxrt-rs Matrix chat for help / feedback / complaints.
- Review the higher-level changes (below) and the HAL API. Testing shows that drivers are no worse than what we had in the 0.4 HAL. I also tried to fix issues like #111. I want to know if you think this is or isn't an improvement, and if we could use this as a foundation for future development. "Future development" includes adding new drivers, defining BSPs, supporting the broader embedded Rust ecosystem, and whatever else you want to do.
- If you like what you see, let me know what you think some next steps are. I have my list below.

I don't expect a line-by-line review right now. After we discuss, I'll clean up the history, and make it easier to review and merge.

Try it out
----------

All examples build and run on the

- Teensy 4 (4.0 and 4.1).
- i.MX RT 1010 EVK.

Select examples also work for the Cortex M7 of the i.MX RT 1170 EVK. This chip support is WIP.

After pulling this branch, you can quickly light your Teensy's LED:

```
cargo objcopy --features=board/teensy4 --example=hal_led --target=thumbv7em-none-eabihf -- -O ihex hal_led.hex
# Load 'hal_led.hex' using your favorite Teensy tool.
```

You can also light your i.MX RT 1010 EVK's LED with `cargo flash`. You need an unreleased version of the tool:

```
cargo install cargo-flash --git=https://github.com/probe-rs/cargo-flash.git
cargo flash --features=board/imxrt1010evk --release --chip=mimxrt1010 --example=hal_led --target=thumbv7em-none-eabihf
```

See `board/README.md` for more information on board support. For tips on adding new boards, see the CONTRIBUTING guide.

What's new?
-----------

[...]

PR #122, opened 2022-12-12
FlexCAN peripheral addition

dstric-aqueduct:
This pull request adds functionality for the imxrt's FlexCAN1 and FlexCAN2 peripherals. 

See corresponding update to teensy4-rs crate: https://github.com/mciantyre/teensy4-rs/pull/124

Credit to Antonio Brewer (tonton81) for much of the API inspiration from the FlexCAN_T4 library: https://github.com/tonton81/FlexCAN_T4


imxrt-hal/src/can/mod.rs
```diff
@@ -0,0 +1,1114 @@
...
+///
+pub struct CAN<M> {
+    reg: ral::can::Instance,
+    _module: PhantomData<M>,
+    source_clock: ccm::Frequency,
+    _mailbox_reader_index: u8,
+}
+
+const CAN1_ADDR: u32 = 0x401d0000;
+const CAN2_ADDR: u32 = 0x401d4000;
```
mciantyre, line 122: Future-proofing consideration in light of #121: CAN1 and CAN2 are at different addresses on (at least) the 1170. It looks like these are only needed for the `base_address()` implementation.

- If we could remove `base_address()`, then we could also remove these two constants.
- If we need to keep `base_address()`, let's acquire the base address from the `imxrt-ral` instance.

    ```rust
    pub fn base_address(&self) -> u32 {
        let addr: *const ral::can::RegisterBlock = &*self.reg;
        addr as u32
    }
    ```

    The `imxrt-ral` instance is (expected to be) correct no matter the target MCU.

mciantyre (commented):
Thanks for putting this together. My first pass is a brief code study. I'm not too familiar with this peripheral, but I'll try to get some time to study the FlexCAN chapter in the reference manual.

I notice that we're re-naming the package and files. I'm guessing that's to support Teensy 4 prototyping. If that's the case, would you mind rebasing this branch onto the `maint-v0.4` branch, then re-targeting the PR to `maint-v0.4` (edit the PR, and change the base branch)? Once this happens, we won't have to re-name all these files, and it will simplify integration and release into the 0.4 HAL.

My review include considerations for #121. Once we're in a good state, I can port this driver into the next HAL.

dstric-aqueduct:
> Thanks for putting this together. My first pass is a brief code study. I'm not too familiar with this peripheral, but I'll try to get some time to study the FlexCAN chapter in the reference manual.
> 
> I notice that we're re-naming the package and files. I'm guessing that's to support Teensy 4 prototyping. If that's the case, would you mind rebasing this branch onto the `maint-v0.4` branch, then re-targeting the PR to `maint-v0.4` (edit the PR, and change the base branch)? Once this happens, we won't have to re-name all these files, and it will simplify integration and release into the 0.4 HAL.
> 
> My review include considerations for #121. Once we're in a good state, I can port this driver into the next HAL.

Hey Ryan - first thanks to you and Tom for all the work on these projects! They're fantastic!

I merged the `maint-v0.4` branch into the `dev/can` branch and then re-targeted to the `maint-v0.4` branch for the PR. I think the new file diffs are only the Can feature additions at this point.

imxrt-hal/src/can/frame.rs
```diff
@@ -0,0 +1,458 @@
...
+
+impl PartialEq for Data {
+    fn eq(&self, other: &Self) -> bool {
+        self.as_ref() == other.as_ref()
+    }
+}
+
+impl Eq for Data {}
+
+#[cfg(feature = "unstable-defmt")]
```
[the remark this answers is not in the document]
dstric-aqueduct: Correct, these are artifacts from the bxcan library - removed in the latest version.

imxrt-hal/src/can/mod.rs
```diff
@@ -0,0 +1,1114 @@
...
+}
+
+#[inline]
+fn constrain<T: PartialOrd>(value: T, min: T, max: T) -> T {
+    match value {
+        v if v < min => min,
+        v if v > max => max,
+        _ => value,
+    }
+}
```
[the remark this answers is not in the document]
dstric-aqueduct: Yes, removed the function and replaced with the [Ordering clamp](core::cmp::Ordering::clamp)[core::cmp::Ordering::clamp](https://doc.rust-lang.org/stable/core/cmp/trait.Ord.html#method.clamp) function.

imxrt-hal/src/can/mod.rs
dstric-aqueduct: The base address is needed to access FlexCAN mailboxes, which are not part of the RAL imxrt-ral crate at this point. I removed the constants and modified the `base_address()` as above.

teburd:
Thanks for submitting this! It's been a requested feature by several people and its great to see it being contributed.

I'd note that the commit set should be rebased, probably squashed a bit, and cleaned up before merging to make looking at the history a bit easier to understand. Commits with one word descriptions aren't going to be all that helpful if we ever need to bisect/blame/etc. Really most of this work I'd think could be squashed down into a couple of commits with a nice commit message describing the flexcan driver being added, how it works, and why.

Happy holidays! Again thank you very much for contributing this.

imxrt-hal/src/can/mod.rs
```diff
@@ -0,0 +1,1114 @@
...
+            2 => {
+                unsafe { modify_reg!(ral::ccm, CCM, CCGR0, |reg| reg | 0x3C0000) };
+            }
+            u => {
+                log::error!("Invalid Can instance (set_ccm_ccg): {:?}", u);
+            }
+        }
+    }
+
+    fn set_tx(&mut self) {
```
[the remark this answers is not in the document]
dstric-aqueduct: Agreed that it would be preferable to stick to the convention: see this fork of [`imxrt-iomuxc`](https://github.com/imxrt-rs/imxrt-iomuxc/compare/master...dstric-aqueduct:imxrt-iomuxc:dev/can) for an implementation of the Pin trait for the CAN1/2 periphs on the 1060.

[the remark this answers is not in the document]
dstric-aqueduct: Commit [`b4eca9f`](https://github.com/imxrt-rs/imxrt-hal/pull/122/commits/b4eca9fedca725de1902de3187b6d0d9166cb453) adopts the convention above.

dstric-aqueduct:
Well...it appears my attempts to squash the fork commit history did not go well as there are now a bunch of file diffs. Ian, Tom - I'll leave it to you on the best way to go about making the PR history clean.

mciantyre (commented):
Here's how I would try to clean up the history. I encourage you to take some time to try this out, and see how it works. This isn't the rebase that Tom and I suggested, but it's a quick way to achieve the same outcome and simplify the history.

1. **`git reset --soft maint-v0.4`**

   After this, the CAN-specific changes (the contents of this diff, as I write this) are in the index. You can see the changes with `git status`. Also, the `HEAD` of `dev/can` is now `maint-0.4`.

2. **`git commit -m "Prototype a FlexCAN peripheral"`**

    I agree with Tom that having more content in the commit message is helpful. I'll leave that to you.

3. **`git push [your-remote-name] dev/can --force`**

    After this, that one new commit from 2 appears in the PR.

imxrt-hal/src/can/frame.rs
```diff
@@ -0,0 +1,451 @@
...
+
+    TxInactive = 0b1000,
+    TxAbort = 0b1001,
+    TxOnce = 0b1100,
+    TxAnswer = 0b1110,
+
+    Unknown,
+}
+
+impl ::core::convert::TryFrom<u8> for FlexCanMailboxCSCode {
```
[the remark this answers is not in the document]
dstric-aqueduct: Yes, this makes sense. Implemented `::core::convert::From<u8>` instead of `::core::convert::TryFrom<u8>` for `FlexCanMailboxCSCode`.

imxrt-hal/src/can/id.rs
```diff
@@ -0,0 +1,126 @@
+/// Standard 11-bit CAN Identifier (`0..=0x7FF`).
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub struct StandardId(u16);
```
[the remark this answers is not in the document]
dstric-aqueduct: Removed in favor of embedded-hal 0.2 StandardId.

imxrt-hal/src/can/id.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// Returns this CAN Identifier as a raw 16-bit integer.
+    #[inline]
+    pub fn as_raw(&self) -> u16 {
+        self.0
+    }
+}
+
+/// Extended 29-bit CAN Identifier (`0..=1FFF_FFFF`).
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub struct ExtendedId(u32);
```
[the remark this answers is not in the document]
dstric-aqueduct: Removed in favor of embedded-hal 0.2 ExtendedId.

imxrt-hal/src/can/mod.rs
```diff
@@ -0,0 +1,998 @@
...
+            17 => Some([5, 7, 6]),
+            18 => Some([6, 7, 6]),
+            19 => Some([6, 7, 7]),
+            20 => Some([7, 7, 7]),
+            _ => None,
+        }
+    }
+
+    pub fn set_baud_rate(&mut self, baud: u32) {
+        let clock_freq = 24_000_000;
```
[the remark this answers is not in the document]
dstric-aqueduct: Yes, using `get_clock` instead of hard coding in baud rate calculation/register config.

mciantyre (changes requested):
Sorry, but I'm pausing my review until we simplify the commit history and change how we apply revisions. I notice that corrections are changing back to their pre-review state. See my latest comment on the `id.rs` file. I'm concerned that this might happen with harder-to-notice, important revisions.

If cleaning up the history is troublesome, I can provide a starting point. I pushed [a `dev/can` branch to the main repository](https://github.com/imxrt-rs/imxrt-hal/tree/dev/can) with a simplified history and a base on `maint-v0.4`. I'll stress that this is just a starting point; I may not have integrated all the changes you expect.

If that branch looks reasonable, I recommend using it for development and review. You can fetch the branch, locally re-name it to `dev/can`, then force-push it over your remote's `dev/can` branch to update this PR.

dstric-aqueduct:
> Sorry, but I'm pausing my review until we simplify the commit history and change how we apply revisions. I notice that corrections are changing back to their pre-review state. See my latest comment on the `id.rs` file. I'm concerned that this might happen with harder-to-notice, important revisions.
> 
> If cleaning up the history is troublesome, I can provide a starting point. I pushed [a `dev/can` branch to the main repository](https://github.com/imxrt-rs/imxrt-hal/tree/dev/can) with a simplified history and a base on `maint-v0.4`. I'll stress that this is just a starting point; I may not have integrated all the changes you expect.
> 
> If that branch looks reasonable, I recommend using it for development and review. You can fetch the branch, locally re-name it to `dev/can`, then force-push it over your remote's `dev/can` branch to update this PR.

Hey Ian - thanks for creating the `dev/can` branch in the main repo. I was able to force-push to the remote's branch so I think we're finally in sync now.

imxrt-hal/src/can/embedded_hal.rs
```diff
@@ -0,0 +1,105 @@
...
+            None => Err(nb::Error::Other(NoDataError { _priv: () })),
+        }
+    }
+}
+
+impl can::Error for NoDataError {
+    fn kind(&self) -> can::ErrorKind {
+        can::ErrorKind::Overrun
+    }
+}
```
mciantyre, line 37: Couple of questions / thoughts about our error handling:

- Can we not detect any other errors in this driver? `can::ErrorKind` has a variety of other errors that could help others with debugging.
- Is `Overrun` the proper kind of error for a "no data error?" My understanding of _overrun_ is "the buffer is full, and you tried to put more data in." This implies that there _is_ data available. If that's the right interpretation, maybe `ErrorKind::Other` would be more appropriate?
- We're using `NoDataError` as the error returns for both CAN transmit and receives (above). Since we need a single type to support multiple operations, should we consider a more generic `Error` enum, possibly non-exhaustive? This would let us incrementally add support for new errors without breaking users.

    ```rust
    /// FlexCAN-specific errors.
    #[non_exhaustive]
    pub enum Error {
      /// Receive buffer is empty, and there's no data.
      NoData,
    }

    impl can::Error for Error { /* ... */ }
    ```

imxrt-hal/src/can/frame.rs
```diff
@@ -0,0 +1,445 @@
...
+    const STANDARD_SHIFT: u32 = 18;
+    const EXTENDED_SHIFT: u32 = 0;
+
+    /// Creates a new standard identifier (11bit, Range: 0..0x7FF)
+    ///
+    /// Panics for IDs outside the allowed range.
+    #[inline(always)]
+    pub fn new(id: u32) -> Self {
+        Self(id)
+    }
```
mciantyre, line 310: > Panics for IDs outside the allowed range.

This constructor doesn't seem to panic. Are the docs out of date, or is this missing a bounds check?

---

What are your thoughts on making `new` a private constructor? This would prevent users from creating an invalid `IdReg`, which might be useful if we intend to have no bounds check.

imxrt-hal/src/can/frame.rs
```diff
@@ -0,0 +1,445 @@
...
+    /// Get the 4 bit code content for a CodeReg.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if a matching [`FlexCanMailboxCSCode`] is not found.
+    #[inline(always)]
+    pub fn code(&self) -> FlexCanMailboxCSCode {
+        FlexCanMailboxCSCode::try_from(((self.0 & Self::CODE_MASK) >> Self::CODE_SHIFT) as u8)
+            .unwrap()
+    }
```
mciantyre, line 247: Now that we implemented `From<u8>` instead of `TryFrom<u8>` for `FlexCanMailboxCSCode`, does this still panic? Could we use `from` and avoid an `unwrap()`?

imxrt-hal/src/can/frame.rs
```diff
@@ -0,0 +1,445 @@
...
+
+    pub fn timestamp(&self) -> u16 {
+        ((self.0 & Self::TIMESTAMP_MASK) >> Self::TIMESTAMP_SHIFT) as u16
+    }
```
mciantyre, line 251: Could we document some properties of this timestamp? Let me know if these sound correct. I'm basing this on the reference manual, mostly the free-running timer register documentation.

```suggestion
    /// The frame timestamp, captured from the peripheral's free-running timer.
    ///
    /// The timestamp represents the time at which a TX or RX frame appears on
    /// the bus. The counter increments at the CAN bus baud rate. The counter pauses
    /// when the driver is frozen, or the processor is in debug mode.
    ///
    /// A frame's timestamp resets to zero when either
    ///
    /// - the counter wraps around at the 16 bit boundary.
    /// - a message is received in mailbox 0. This happens if time sync is enabled,
    ///   and if the message passed the mailbox filter.
    ///
    /// Users may also override the timestamp to any specific value.
    pub fn timestamp(&self) -> u16 {
        ((self.0 & Self::TIMESTAMP_MASK) >> Self::TIMESTAMP_SHIFT) as u16
    }
```

imxrt-hal/src/can/mod.rs
```diff
@@ -0,0 +1,992 @@
...
+
+    pub fn is_can1(&self) -> bool {
+        M::USIZE == 1
+    }
+
+    pub fn is_can2(&self) -> bool {
+        M::USIZE == 2
+    }
+
+    pub fn base_address(&self) -> u32 {
```
mciantyre, line 250: Could we make this private or `pub(crate)`? Or, why might the user need access to the base address?

imxrt-hal/Cargo.toml
```diff
@@ -22,9 +22,9 @@ log = "0.4.8"
 rand_core = { version = "0.5", default-features = false, optional = true }
 
 [dependencies.imxrt-iomuxc]
-version = "0.1.2"
-git = "https://github.com/imxrt-rs/imxrt-iomuxc.git"
-branch = "v0.1"
+features = ["imxrt106x"]
+git = "https://github.com/dstric-aqueduct/imxrt-iomuxc.git"
+tag = "v0.1.5-can"
```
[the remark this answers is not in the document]
dstric-aqueduct: Done - PR [here](https://github.com/imxrt-rs/imxrt-iomuxc/pull/29).

imxrt-hal/Cargo.toml
```diff
@@ -34,7 +34,7 @@ features = ["device"]
 bench = false
 
 [features]
-default = ["embedded-hal/unproven"] # Allows us to access the new digital pin traits
+default = ["embedded-hal/unproven", "imxrt1062"] # Allows us to access the new digital pin traits
```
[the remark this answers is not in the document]
dstric-aqueduct: Yes, for prototyping and I will remove it.

imxrt-hal/src/can/embedded_hal.rs
dstric-aqueduct: > Couple of questions / thoughts about our error handling:
> 
> * Can we not detect any other errors in this driver? `can::ErrorKind` has a variety of other errors that could help others with debugging.

Yes, that would be helpful - I will need to dig in a bit to the implementation to see where the other `ErrorKind`s may fit. 

> * ~Is `Overrun` the proper kind of error for a "no data error?" My understanding of _overrun_ is "the buffer is full, and you tried to put more data in." This implies that there _is_ data available. If that's the right interpretation, maybe `ErrorKind::Other` would be more appropriate?~ My understanding and interpretation may be wrong; `embedded_hal::can::ErrorKind::Overrun` is specifically for the _receive_ buffer. I suppose I'm not sure if `Overrun` is intended to tell the user "I received so much data from another device that I overran my buffer," or if it tells the user "there's no data in the receive buffer."

Yes, that is my interpretation too and I agree the naming is misleading. Updated to:

`can/mod.rs`

```rust
`/// Error from the FlexCan peripheral.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum Error {
/// Error indicating that no received data is 
/// available in the mailboxes
NoRxData,
/// Error indicating that no transmit mailboxes
/// are available
NoTxMailbox, 
/// A wrapper around the [`embedded_hal::ErrorKind`]
/// enum
EmbeddedHal(embedded_hal::ErrorKind),
}
```
```rust
#[inline(always)]
pub fn transmit(&mut self, frame: &Frame) -> nb::Result<(), Error> {
    for i in self.mailbox_offset()..self.get_max_mailbox() {
        if let FlexCanMailboxCSCode::TxInactive =
            FlexCanMailboxCSCode::from_code_reg(self.read_mailbox_code(i))
        {
            self.write_tx_mailbox(i, frame);
            return Ok(());
        }
    }
    Err(nb::Error::Other(Error::NoTxMailbox))
}
```

`can/embedded_hal.rs`

```rust
impl<M> can::Can for CAN<M>
where
    M: Unsigned,
{
    type Frame = Frame;

    type Error = Error;

    fn transmit(&mut self, frame: &Self::Frame) -> nb::Result<Option<Self::Frame>, Self::Error> {
        match self.transmit(frame) {
            Ok(_status) => Ok(Some(frame.clone())),
            Err(nb::Error::WouldBlock) => Err(nb::Error::WouldBlock),
            Err(nb::Error::Other(e)) => Err(nb::Error::Other(e)),
        }
    }

    fn receive(&mut self) -> nb::Result<Self::Frame, Self::Error> {
        match self.read_mailboxes() {
            Some(d) => Ok(d.frame),
            None => Err(nb::Error::Other(Error::NoRxData)),
        }
    }
}

impl can::Error for Error {
    fn kind(&self) -> can::ErrorKind {
        match self {
            Self::NoRxData => can::ErrorKind::Other,
            Self::NoTxMailbox => can::ErrorKind::Other,
            Self::EmbeddedHal(e) => e.kind()
        }
    }
}
```

> * We're using `NoDataError` as the error returns for both CAN transmit and receives (above). Since we need a single type to support multiple operations, should we consider a more generic `Error` enum, possibly non-exhaustive? This would let us incrementally add support for new errors without breaking users.
>   ```rust
>   /// FlexCAN-specific errors.
>   #[non_exhaustive]
>   pub enum Error {
>     /// Receive buffer is empty, and there's no data.
>     NoData,
>   }
>   
>   impl can::Error for Error { /* ... */ }
>   ```

imxrt-hal/src/can/frame.rs
dstric-aqueduct: > > Panics for IDs outside the allowed range.
> 
> This constructor doesn't seem to panic. Are the docs out of date, or is this missing a bounds check?
> 

Updated to:

```rust
/// Creates a new standard identifier (11bit, Range: 0..=0x7FF)
///
/// Panics for IDs outside the allowed range.
#[inline(always)]
pub fn new_standard(id: StandardId) -> Self {
    assert!(id.as_raw() <= StandardId::MAX.as_raw());
    let id = u32::from(id.as_raw()) << Self::STANDARD_SHIFT;
    Self::new(id)
}

/// Creates a new extendended identifier (29bit , Range: 0..=0x1FFFFFFF).
///
/// Panics for IDs outside the allowed range.
#[inline(always)]
pub fn new_extended(id: ExtendedId) -> Self {
    assert!(id.as_raw() <= ExtendedId::MAX.as_raw());
    let id = u32::from(id.as_raw()) << Self::EXTENDED_SHIFT;
    Self::new(id)
}
```

> What are your thoughts on making `new` a private constructor? This would prevent users from creating an invalid `IdReg`, which might be useful if we intend to have no bounds check.

Yes, no issues making `new` private.

imxrt-hal/src/can/frame.rs
dstric-aqueduct: Your documentation looks good to me and has been added.

imxrt-hal/src/can/mod.rs
dstric-aqueduct: Yes, `pub(crate)` works here. No access to the base address should be required by the user.

imxrt-hal/src/can/mod.rs
```diff
@@ -0,0 +1,1040 @@
...
+            if self._mailbox_reader_index >= max_mailbox {
+                self._mailbox_reader_index = self.mailbox_offset();
+                cycle_limit -= 1;
+                if cycle_limit == 0 {
+                    return None;
+                }
+            }
+            if (self.read_imask() & (1_u64 << self._mailbox_reader_index)) != 0 {
+                self._mailbox_reader_index += 1;
+                continue; /* don't read interrupt enabled mailboxes */
```
mciantyre, line 988: > don't read interrupt enabled mailboxes

Is this documented somewhere for the user? Although `handle_interrupt` also returns a `MailboxData`, it might not be obvious that `handle_interrupt` is XOR `read_mailboxes`.

imxrt-hal/src/can/mod.rs
```diff
@@ -0,0 +1,994 @@
...
+                        };
+                        this.write_mailbox(i, Some(code), None, None);
+                        let eacen = read_reg!(ral::can, this.reg, CTRL2, EACEN == EACEN_1);
+                        let rximr = 0_32 | {
+                            if eacen {
+                                1_u32 << 30
+                            } else {
+                                0
+                            }
+                        };
```
mciantyre, line 514: `clippy` finds `0_32` curious. Is `0_32` intended to be `0u32` or `32`? If it's just a zero, it looks like we could remove the bitwise OR.

```rust
let rximr: u32 = if eacen { 1 << 30 } else { 0 };
```

imxrt-hal/src/can/mod.rs
```diff
@@ -0,0 +1,994 @@
...
+        ral::read_reg!(ral::can, self.reg, MCR, RFEN == RFEN_1)
+    }
+
+    fn mailbox_offset(&self) -> u8 {
+        if self.fifo_enabled() {
+            let max_mailbox = self.get_max_mailbox() as u32;
+            let num_rx_fifo_filters = (read_reg!(ral::can, self.reg, CTRL2, RFFN) + 1) * 2;
+            let remaining_mailboxes = max_mailbox - 6_u32 - num_rx_fifo_filters;
+            /* return offset MB position after FIFO area */
+            if max_mailbox < max_mailbox - remaining_mailboxes {
```
mciantyre, line 722: clippy impressed me with this one. Is there a different way to express this mailbox offset?

```
warning: you are trying to use classic C underflow conditions that will fail in Rust
   --> imxrt-hal/src/can/mod.rs:722:16
    |
722 |             if max_mailbox < max_mailbox - remaining_mailboxes {
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#overflow_check_conditional
    = note: `#[warn(clippy::overflow_check_conditional)]` on by default
```

imxrt-hal/src/can/mod.rs
```diff
@@ -0,0 +1,994 @@
...
+
+    fn fifo_enabled(&self) -> bool {
+        ral::read_reg!(ral::can, self.reg, MCR, RFEN == RFEN_1)
+    }
+
+    fn mailbox_offset(&self) -> u8 {
+        if self.fifo_enabled() {
+            let max_mailbox = self.get_max_mailbox() as u32;
+            let num_rx_fifo_filters = (read_reg!(ral::can, self.reg, CTRL2, RFFN) + 1) * 2;
+            let remaining_mailboxes = max_mailbox - 6_u32 - num_rx_fifo_filters;
```
mciantyre, line 720: Is this `6_u32` the same as `NUMBER_FIFO_RX_MAILBOXES`?

mciantyre (commented):
I rebased the branch onto `maint-v0.4` to isolate clippy warnings, formatted the code to pass the next CI step, and collapsed into one commit.

I appreciate help addressing the 21 clippy warnings. You can reproduce them locally by checking out the updated branch, then running clippy like

```
cargo clippy --features=imxrt1062
```

If I'm left to fix these, I'll need clarity on the intent of some code; see inline comments.

imxrt-hal/src/can/mod.rs
```diff
@@ -0,0 +1,972 @@
...
+                result = clock_freq / baud / (divisor + 1);
+                if result <= 25 {
+                    error = (baud - (clock_freq / (result * (divisor + 1)))) as i16;
+                    if error < 0 {
+                        error *= -1;
+                    }
+                    if error < best_error {
+                        best_error = error;
+                        best_divisor = divisor;
+                    }
```
mciantyre, line 377: Unsigned subtraction of `baud` and `clock_freq / [...]` -- before casting to `i16` -- can overflow. This can cause a panic when building with default _debug_ configurations.

```
[ERROR teensy4_panic]: panicked at 'attempt to subtract with overflow', /path/to/imxrt-hal/src/can/mod.rs:370:29
```

I can reproduce the panic in the teensy4-rs `can.rs` example (as of [this commit](https://github.com/mciantyre/teensy4-rs/pull/124/commits/d176a6d17ea2c7e022aad9d21461168d3cc59473)). The snippet below shows how I'm building the example. I haven't stepped through the code to understand what specific iteration produces this panic.

```
cargo build --target=thumbv7em-none-eabihf --example=can --all-features
```

My first thought is to use something like [`abs_diff`](https://doc.rust-lang.org/std/primitive.u32.html#method.abs_diff). However, if the absolute difference is larger than what fits in an `i16`, we'll truncate that difference once we cast from `u32` to `i16`.
- Could we revise this routine so that it's less likely to panic in debug builds?
- Are the truncating casts to `i16` necessary?

imxrt-hal/src/can/mod.rs
```diff
@@ -0,0 +1,972 @@
...
+//!
+//! let mut can1 = can1_builder.build(
+//!     peripherals.iomuxc.ad_b1.p08,
+//!     peripherals.iomuxc.ad_b1.p09
+//! );
+//!
+//! can1.set_baud_rate(1_000_000);
+//! can1.set_max_mailbox(16);
+//! can1.enable_fifo();
+//! can1.set_fifo_interrupt(true);
```
mciantyre, line 27: Question about this API and behavior. I'm effectively testing this documentation example through the `can.rs` example proposed in the teensy4-rs repo (as of [this commit](https://github.com/mciantyre/teensy4-rs/pull/124/commits/d176a6d17ea2c7e022aad9d21461168d3cc59473)).

1. I'm changing this driver configuration to `false`, and keeping all other `can1` calls in the block the same.

    ```rust
    can1.set_fifo_interrupt(false);
    ```

2. I'm then unmasking the CAN1 interrupt before the LED-toggling-CAN-I/O loop.

    ```rust
    // NEW:
    unsafe { cortex_m::peripheral::NVIC::unmask(bsp::interrupt::CAN1) };
    loop { /* Delay, toggle LED, read CAN mailboxes, send CAN frame */ }
    ```

3. I implementing the CAN1 ISR elsewhere in the example. It looks like

    ```rust
    use bsp::interrupt;
    #[cortex_m_rt::interrupt]
    fn CAN1() {
        log::warn!("CAN1 interrupt fired");
    }
    ```

I'm expecting step 1 to disable the CAN interrupts; that seems to be the only public method with something implying interrupt activation in the name. So even though I'm unmasking the CAN1 interrupt, I'm never expecting the ISR to activate.

But when I actually build and run this, I see that the CAN1 ISR is firing repeatedly. I'm connecting to the USB serial console to see all the log warnings.

Is there something else in the driver that's enabling interrupts, and can it be changed by the user? Or, am I using this driver incorrectly?

imxrt-hal/src/can/mod.rs
```diff
@@ -0,0 +1,1040 @@
...
+                    mailbox_number,
+                })
+            }
+            _ => None,
+        }
+    }
+
+    /// Write the registers of a mailbox.
+    #[inline(always)]
+    fn write_mailbox(
```
myCatPlaysFetch, line 794: There are FlexCAN errata, at least on the RT1064, that require some extra precautions and/or workarounds. This function and elsewhere in the module would be affected. Those are 5641 (not on RT1064), 5829, 6032, and 9595. Should this driver be expected to know of and handle these errata? Or would there be more board-specific drivers for that? Overall excellent work, just curious how something like that is handled.

mciantyre: Thanks for helping with the review. If an erratum's workaround can act as the driver's default behavior, that would be ideal, and we should handle it here. This cuts down on configurations / code paths we maintain. But, if we need separate code paths to handle a driver's errata, we could consider crate configurations / separate driver crates / common- and chip-specific HAL combinations.

imxrt-hal/src/can/mod.rs
dstric-aqueduct: This has been updated to use the `abs_diff` method.

imxrt-hal/src/can/mod.rs
dstric-aqueduct: From the [manual](https://www.pjrc.com/teensy/IMXRT1060RM_rev2.pdf) (pg 2553): 

_The module can generate up to 70 interrupt sources (64 interrupts due to message buffers
and 6 interrupts due to Ored interrupts from MBs, Bus Off, Error, Tx Warning, Rx
Warning and Wake Up)).

If the Rx FIFO is enabled (bit RFEN on MCR set), the interrupts corresponding to MBs 0
to 7 have a different behavior. Bit 7 of the IFLAG1 becomes the "FIFO Overflow" flag;
bit 6 becomes the FIFO Warning flag, bit 5 becomes the "Frames Available in FIFO flag"
and bits 4-0 are unused. See Interrupt Flags 1 Register (FLEXCAN_IFLAG1) for more
information.

A combined interrupt for all MBs is also generated by an Or of all the interrupt sources
from MBs. This interrupt gets generated when any of the Mailboxes or FIFO generates an
interrupt. The Arm must read the IFLAG Registers to determine which MB or FIFO
caused the interrupt.

The other 5 interrupt sources (Bus Off, Error, Tx Warning, Rx Warning and Wake Up
generate interrupts like the MB ones, and can be read from both the Error and Status
Register 1 and 2. The Bus Off, Error, Tx Warning and Rx Warning interrupt mask bits
are located in the Control 1 Register and the Wake-Up interrupt mask bit is located in the
MCR._

This method is currently only setting the BUF5I bit in IMASK1 register to interrupt when a frame is available (Frames Available in FIFO flag...):

_If the Rx FIFO is not enabled, this bit flags the interrupt for MB5. If the Rx FIFO is enabled, this flag
indicates that at least one frame is available to be read from the Rx FIFO.
This flag is cleared by the FlexCAN whenever the bit MCR[RFEN] is changed by Arm writes_

I agree this could be more transparent. More importantly, masking the interrupt vector for the interrupt reason would be beneficial. 

This would let us debug why the interrupt is always firing for you (likely Bus Off with no transceiver attached.)

mciantyre: When I run the `can.rs` example in the teensy4-bsp project, it logs a few control registers before entering its I/O loop. Here's three register values I see, followed by a discussion of interrupt-specific register fields.

```
[INFO imxrt_hal::can]: MCR: 64EB300F
```

In `MCR`,

- `WRNEN` is set. However, we're never setting `CTRL1[TWRNMSK]` or `CTRL1[RWRNMSK]`, so (from what I can tell) we shouldn't observe an interrupt when error counters cross their threshold. If that's correct, I'm not sure we need to set `WRNEN`.
- `WAKMSK` and `SLFWAK` are set. This means that we could trigger an interrupt when exiting doze or stop modes. (This definitely isn't happening in my limited testing, but it may be a non-obvious behavior of this driver.)

```
[INFO imxrt_hal::can]: CTRL1: 1624002
```

In `CTRL1`,

- `BOFFMSK` is clear, so there's no "bus off" interrupt.
- `ERRMSK` is set, so we generate an interrupt when `ESR1[ERRINT]` is high. This is a hasty observation, but a project-wide CTRL+F for "[ESR1](https://docs.rs/imxrt-ral/0.4.4/imxrt_ral/imxrt106/peripherals/can/ESR1/index.html)" shows no hits. This could indicate we're not clearing error interrupts.
- `TWRNMSK` and `RWRNMSK` are both clear; see `MCR[WRNEN]` discussion, above.

```
[INFO imxrt_hal::can]: CTRL2: 870000
```

In `CTRL2`,

- `BOFFDONEMSK` is clear, so there's no "bus off done" interrupt.

I'm not evaluating any message buffer interrupts right now. Nevertheless, I'm wary of `ERRMSK` being set, especially if we're not handling any error and status flags.

collin80:
The underlying hardware for the Teensy 4/4.1/Micromod supports 3 CAN buses, not two. Also, the third hardware bus is CAN-FD capable. It seems that currently your HAL doesn't have any concept of this third CAN bus. However, would this be possible to add? I realize that the addition of CAN-FD may complicate things. But, it is physically present on the chip so for teensy4-rs it would seem proper for it to eventually be supported.

I'm not entirely asking that "someone else do it for me" but kind of. I am toward the beginning of my rust journey so currently it may be difficult for me to take on such a thing by myself. However, I'm not entirely green at making CAN libraries. I use Tony's FlexCAN-T4 library on Teensy Micromod but on other chips I used my own CAN libraries. So, if this may be something I could potentially help to do I'm open to pointers as to how to get started.

mciantyre:
#171 rebases this driver onto a later commit.

PR #124, opened 2023-02-26
Integrate i.MX RT 1060 EVK board for examples

mciantyre:
All examples build. However, I'm not equipped with a 1060 EVK, so I haven't tested any of these on hardware.

Pinouts correspond to board revision A3. Looks like board revision B1 changes some header identifiers (J24 on A3 is J17 on B1, for example). I haven't done a diligent comparison to see if any pin routing is different.

Depends on imxrt-rs/imxrt-boot-gen#10. Once that's merged and released, I'll remove the git dependency here.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -84,9 +84,11 @@ jobs:
         # Boards that support all examples, and should build with additional drivers.
         - --examples --features=board/teensy4,board/lcd1602
         - --examples --features=board/imxrt1010evk,board/lcd1602
+        - --examples --features=board/imxrt1060evk,board/lcd1602
         # SPI examples (might break other examples)
         - --example=hal_spi --example=rtic_spi --example=async_dma_spi --features=board/teensy4,board/spi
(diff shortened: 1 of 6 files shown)


mciantyre:
Here's (what should be) the command to build and flash the `rtic_logging` example.

```
cargo install cargo-flash
cargo flash --features=board/imxrt1060evk --release --chip=mimxrt1060 --example=rtic_logging --target=thumbv7em-none-eabihf
```

You may need to press the reset button after programming. After programming, you should see the LED blinking. If you connect to the DAP's serial port, you should see logging output.

I'll keep this open for a week for anyone to provide feedback.

teburd (approved):

mciantyre:
Debugged and tested this over Matrix. After correcting some clock gates, we observed a blinking LED.

> Pinouts correspond to board revision A3. Looks like board revision B1 changes some header identifiers [...]

The schematic for revision B1 has a revision history on the first page. The entry for rev code B indicates that I2C3 SCL and SDA are swapped, differing from rev A boards. We're not using that peripheral at the moment, but just a heads up in case that changes. There are other changes listed in the revision notes that I haven't evaluated.

PR #125, opened 2023-02-26
Add 1020 chip configuration

Florob:
Add chip configuration for the i.MX RT 1020 in accordance with the RM. Basic testing was done on EVK as well as an own design.

There is one TODO comment in here: In `imxrt1010.rs` `Clko2Selection` currently calls the`trace_clk_root` `TracClk` instead of `TraceClk`. I'm not sure if that is a typo or intentional. And regardless of whether it is how to spell it for 1020.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -55,6 +55,7 @@ jobs:
       matrix:
         chips:
         - imxrt-ral/imxrt1011,imxrt1010
+        - imxrt-ral/imxrt1021,imxrt1020
         - imxrt-ral/imxrt1061,imxrt1060
         - imxrt-ral/imxrt1062,imxrt1060
         - imxrt-ral/imxrt1064,imxrt1060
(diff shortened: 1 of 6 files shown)


mciantyre (approved):
Awesome, thanks for your continued 1020 contributions. I don't have hardware to play with this, but I checked all code changes against the reference manual. This will be in the next 0.5 release; give me a few days.

> There is one TODO comment in here: In `imxrt1010.rs` `Clko2Selection` currently calls the`trace_clk_root` `TracClk` instead of `TraceClk`. I'm not sure if that is a typo or intentional.

Good find; I typo-ed that. I might include e9ff8f6092028dbed3c5ce7846d5b42f5a451e7d in the 0.5 series, and simultaneously fix the spelling before 0.6. In any case, we'll keep the correct spelling proposed here. I'll remove the TODO comment, add some CI testing, and include a changelog entry before merging.

Florob:
> Awesome, thanks for your continued 1020 contributions. I don't have hardware to play with this, but I checked all code changes against the reference manual. This will be in the next 0.5 release; give me a few days.

Not at all. Thank you for maintaining this project. I'm doing this for a personal project (hence the largest i.MX RT that still has leads) and it would not be feasible without the existing ecosystem.

> Good find; I typo-ed that. I might include https://github.com/imxrt-rs/imxrt-hal/commit/e9ff8f6092028dbed3c5ce7846d5b42f5a451e7d in the 0.5 series, and simultaneously fix the spelling before 0.6.

An associated constant looks like a clever solution :+1:. Thank you for considering this.

PR #134, opened 2023-04-13
Fix #133

### LINKED ISSUE #133: LPSPI transfer not reading data back into buffer
We are currently having an issue where the LPSPI transfer reads no data back into the buffer. Our code worked fine with v0.4.2. Note that the SPI loopback example will obviously appear to work -- even if you disconnect sdo from sdi -- since the original output buffer is never written to.

SploxFox:
This PR fixes #133 where the LPSPI code wasn't writing back to the buffer. Tested on Teensy 4.1 (i.MX RT1062)

### DIFF
--- a/src/common/lpspi.rs
+++ b/src/common/lpspi.rs
@@ -634,9 +634,18 @@ impl<P, const N: u8> Lpspi<P, N> {
         let mut tx_idx = 0usize;
         let mut rx_idx = 0usize;
 
-        while tx_idx < buffer.len() && rx_idx < buffer.len() {
+        // Continue looping while there is either tx OR rx remaining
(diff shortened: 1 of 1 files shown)


mciantyre (approved):
Looks great. Thanks for fixing the loop and working around the hardware.

Now that we're terminating the continuous transfer within the loop, it looks like we can remove the extra transmit command after the loop. I queued that up in https://github.com/imxrt-rs/imxrt-hal/commit/3e1f2aabb50c38d0dcfc21fd32d68d249a6e7b94, and that's what I tested. I'll leave this open for a few more days to answer questions about that removal. After that, I'll release it in 0.5.2

PR #135, opened 2023-04-19
LPSPI sampling configuration API

dstric-aqueduct:
Expose a public method to allow users to change the sampling point of an LPSPI peripheral.

I've experienced RX FIFO overruns when the SAMPLE: SAMPLE_1 (delayed edge) is used with non-hardware controlled chip select pins (ie, manually driving a CS pin with GPIO `embedded_hal` traits). Changing this configuration setting to SAMPLE_0 fixed the issue.

Tested on a Teensy 4.0 and 4.1 with a Trinamic TMC5130 stepper driver IC.  

When using CS0 (pin 10 on Teensy 4.X), the  `SAMPLE_1` setting works fine. When using pin 4, the `SAMPLE_0` must be used to avoid overruns. I'm curious if anyone else has experienced this issue.

### DIFF
--- a/src/common/lpspi.rs
+++ b/src/common/lpspi.rs
@@ -507,7 +507,7 @@ impl<P, const N: u8> Lpspi<P, N> {
     /// When set to `SamplePoint::DelayedEdge`, the LPSPI master will sample the input data
     /// on a delayed LPSPI_SCK edge, which improves the setup time when sampling data.
     ///
-    /// On [`init`](`Lpspi::init`), the `SamplePoint::DelayedEdge` congfiguration is set.
+    /// This driver sets `SamplePoint::DelayedEdge` by default.
     ///
     /// When using a non-hardware controlled chip select pin, the `SamplePoint::DelayedEdge` can lead to
     /// rx fifo overruns, so the `SamplePoint::Edge` configuration may be preferable.
--- a/src/common/lpspi.rs
+++ b/src/common/lpspi.rs
@@ -502,25 +502,6 @@ impl<P, const N: u8> Lpspi<P, N> {
         Interrupts::from_bits_truncate(ral::read_reg!(ral::lpspi, self.lpspi, IER))
     }
 
-    /// Set the sampling point of the LPSPI in master mode.
-    ///
-    /// When set to `SamplePoint::DelayedEdge`, the LPSPI master will sample the input data
-    /// on a delayed LPSPI_SCK edge, which improves the setup time when sampling data.
-    ///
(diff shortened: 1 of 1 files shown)


mciantyre (changes requested):
New API and docs LGTM. However, the method `set_sample_point` should be attached to [`Disabled`](https://github.com/imxrt-rs/imxrt-hal/blob/8e254826866d77e080979a53bf6d956c9941cbb4/src/common/lpspi.rs#L891). From the 1060 reference manual (rev 3 section 48.5.1.9.2),

> CFGR1 should only be written when LPSPI is disabled.

---

We've included the "delayed edge by default" behavior since this driver was originally integrated. Now that the user can change the setting themselves, we could remove that implicit behavior and keep the hardware default. That's a future change for a breaking release.

> When using CS0 (pin 10 on Teensy 4.X), the  SAMPLE_1 setting works fine. When using pin 4, the SAMPLE_0 must be used to avoid overruns. I'm curious if anyone else has experienced this issue.

Sorry, not sure why this sampling point setting would be based on chip select. But if this is the difference between hardware- and software-controlled chip selects, I'm curious if there's a timing issue between CS changes and the LPSPI transactions.

src/common/lpspi.rs
```diff
@@ -493,6 +502,25 @@ impl<P, const N: u8> Lpspi<P, N> {
         Interrupts::from_bits_truncate(ral::read_reg!(ral::lpspi, self.lpspi, IER))
     }
 
+    /// Set the sampling point of the LPSPI in master mode.
+    ///
+    /// When set to `SamplePoint::DelayedEdge`, the LPSPI master will sample the input data
+    /// on a delayed LPSPI_SCK edge, which improves the setup time when sampling data.
+    ///
+    /// On [`init`](`Lpspi::init`), the `SamplePoint::DelayedEdge` congfiguration is set.
```
mciantyre, line 510:
```suggestion
    /// This driver sets `SamplePoint::DelayedEdge` by default.
```

mciantyre: We could also attach this note to the documentation of [`new`](https://github.com/imxrt-rs/imxrt-hal/blob/8e254826866d77e080979a53bf6d956c9941cbb4/src/common/lpspi.rs#L369-L371), describing it with the SPI mode and clock settings.

dstric-aqueduct:
Whoops - accidentally closed when responding - Ian, can you reopen?

mciantyre:
I squashed these commits, added a changelog entry, and merged. If there's any follow-on work between the close and now, we can discuss in a new PR.

PR #140, opened 2023-09-10
Add 10xx FlexIO CCM configuration

Florob:
This adds support for setting source clock and dividers for FlexIO clock roots in accordance with the RMs

This is WIP because I haven't gotten around to testing the current version on actual hardware yet, but I'd like feedback on the general approach.
The issue here is that FLEXIO1_CLK_ROOT on 1010/1020 is FLEXIO2_CLK_ROOT on 106x, so there is some magic involved in getting the module names and doc comments right. Due to this it might also make sense to move this into a separate module and re-export it.

### DIFF
--- a/src/chip/imxrt10xx.rs
+++ b/src/chip/imxrt10xx.rs
@@ -3,6 +3,7 @@
 //! Shared modules may rely on configurations from the `config` module.
 
 pub mod adc;
+#[macro_use]
 pub mod ccm;
 pub mod dcdc;
 #[path = "dma.rs"]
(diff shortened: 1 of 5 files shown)


Florob:
I've finally gotten around to verifying this. After cursing at the PFD frequencies being wildly different from what is in the datasheet for a while I'm now reasonably confident that this is correct.

@mciantyre I agree with your changes and squashed them into the initial commit, as I don't think having them separate makes much sense.

mciantyre (approved):
Thanks for adding this. I rebased onto `main`, and I'll release in 0.5.4.

PR #141, opened 2023-10-23
Use defmt-rtt and panic-probe for evk boards

SpinFast:
The evk boards have a built in debug probe. When running firmware with probe-rs defmt-rtt provides a very fast channel of low cost debug println's that everyone finds really helpful.

Using panic-probe with defmt-rtt has a second benefit in allowing for programs to inform probe-rs the firmware has exited cleanly. This can be used for writing samples and tests that exit.

This also makes the imxrt-log defmt feature optional to the boards otherwise the two implementations of the defmt logger cause duplicate symbols to show up.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -4,4 +4,5 @@ rustflags = [
     "-C", "link-arg=-Tdefmt.x",
     "-C", "link-arg=-Tdevice.x",
     "-C", "link-arg=-error-limit=0",
+    "-C", "link-arg=-nmagic",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,7 +89,7 @@ members = [
 imxrt-dma = "0.1"
 imxrt-iomuxc = "0.2.1"
 imxrt-hal = { version = "0.5", path = "." }
-imxrt-log = { path = "logging" }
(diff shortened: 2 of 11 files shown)


mciantyre:
Happy to merge this or something similar. Ping me if integrating becomes troublesome, and I can help out.

> This also makes the imxrt-log defmt feature optional to the boards otherwise the two implementations of the defmt logger cause duplicate symbols to show up.

I'm OK with this, but this breaking change might require extra work to integrate. Another approach is to disable all default imxrt-log features in `board/Cargo.toml`, then selectively enable `log` and `defmt` depending on the board.

I'm also OK with chucking imxrt-log in favor of defmt-rtt throughout. Teensy 4 users would need to fend for themselves.

SpinFast:
Believe I've fixed the remaining CI issues. I don't have an 1170evk to test on. I tried a few variations on not removing defmt as a default feature for imxrt-log but ran into strange issues with cargo workspaces and features.

SpinFast:
This passes CI now and doesn't remove the defmt from the default features of imxrt-log

mciantyre (approved):
LGTM. Thanks for working through the integration. I tested on a 1170EVK and Teensy 4. There's a few extra commits to use defmt-rtt on the 1170EVK, and one more to ensure that we can still use the defmt logging backend for the Teensy 4.

PR #145, opened 2023-11-18
LPSPI: embedded-hal 1.0 rework

Finomnis:


examples/rtic_spi.rs
```diff
@@ -4,64 +4,108 @@
...
+    async fn demo(cx: demo::Context) {
+        let demo::LocalResources { spi_device, .. } = cx.local;
+
+        loop {
+            Systick::delay(1000.millis()).await;
+
+            // To demonstrate normal operation
+            spi_device
+                .transaction(&mut [
+                    Operation::DelayUs(100),
```
teburd, line 92: Won't these sorts of delays effectively imply a timer queue and async executor exists?

teburd (commented):
I briefly looked this over, the delays in the SPI transaction look very strange, is there a particular reason those are there? Won't this imply the need for a timer queue or scheduler of some kind?

That seems like a pretty big ask of what amounts to a hal library?

Finomnis:
> I briefly looked this over, the delays in the SPI transaction look very strange, is there a particular reason those are there? Won't this imply the need for a timer queue or scheduler of some kind?
> 
> That seems like a pretty big ask of what amounts to a hal library?

I'm uncertain what you mean, to be honest. Which delays?

mciantyre (commented):
Thanks for working through this, and sorry for the slow review. There's lots of work in this branch, including changes to CI tools and updated dependencies. If you think these are helpful to include separately, I can take a look and help with separation.

---

We're free to break APIs and explore new drivers. But I'll admit: I'm going to miss the 0.5 `Lpspi` driver. It wasn't perfect, but I appreciated that the driver exposed the lower-level details of the peripheral to users. I've used this to drive I/O from interrupts (without `async` conveniences). And although I hadn't tried, I thought it could be the foundation for a future `async` driver, either in this package or another package.

During your prototyping, did you hit limitations of the existing `Lpspi` driver that prevented it from being usable as this driver's foundation? (I may be in the minority about the "give users low-level control" design decisions, so I'm happy to scrap it.)

---

Our approach is a single `Lpspi` driver that supports blocking, interrupt-driven `async`, and DMA-driven `async` IO in one / both directions. Having the interrpt and DMA `async` IO in a single implementation lets the user decide if the want interrupts to send data and DMA to receive data, DMA to send and receive data, etc. I'm curious if there's a way to let the user compose these features themselves, without it all being in one larger driver. Here's a rough outline of what it might look like:

<details>
<summary>Outline, trade-offs</summary>

- `lpspi::InterruptTransmit` for asynchronously sending data with interrupts. Includes its own `InterruptHandler`-style object, which is used in the LPSPI ISR. Doesn't care about embedded-hal-async (EHA).
- `lpspi::InterruptReceive` for async receiving data with interrupts. Includes its own `InterruptHandler`-style object, used inside the LPSPI ISR. Doesn't care about EHA.
- `lpspi::DmaTransmit` for async-sending data with DMA. Includes an `InterruptHandler`-style object, used inside a DMA ISR. Doesn't care about EHA.
- `lpspi::DmaReceive`: async-receive, complementary handler used in DMA ISR, no EHA care.
- `lpspi::AsyncLpspi` accepts a combo of the `*Transmit` and `*Receive` splits. The user is already juggling the `InterruptHandler`s, and that's how we realize the async IO. This combiner implements the EHA async traits. The user constructs this.
- Some internal means of splitting the LPSPI instance to realize these APIs. *Waves hands waves hands.*
- Keep blocking behaviors separate.

The user manages two `InterruptHandler` objects, instead of the single `InterruptHandler` shown today. `AsyncLpspi` supports `u16` IO only when composed of the two `Interrupt*` halves. This approach might not let us realize some kinds of optimizations ("use interrupts, not DMA, when you can saturate the TX FIFO and quickly return to caller") without planning.

</details>

My goal with this design would be to build a solid interrupt-driven `async` foundation, then make it easy to eventually develop DMA-driven `async` without changing the large LPSPI driver's plumbing. It looks like we're approaching this idea with the internal read- and write-halves; one more cut, and we're nearly there.

src/common/lpspi/status_watcher.rs
```diff
@@ -0,0 +1,325 @@
...
+
+struct StatusWatcherInner<const N: u8> {
+    transfer_complete_happened: bool,
+    transfer_complete_waker: Option<Waker>,
+    error_caught: Option<LpspiError>,
+    error_caught_waker: Option<Waker>,
+    tx_fifo_watermark_busy: bool,
+    rx_fifo_watermark_busy: bool,
+    tx_fifo_watermark_waker: Option<Waker>,
+    rx_fifo_watermark_waker: Option<Waker>,
```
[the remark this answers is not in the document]
Finomnis: That might actually be true. Haven't thought of it like this yet. It must be guaranteed though, otherwise this falls back to busy waiting.

Finomnis:
> I may be in the minority about the "give users low-level control" design decisions

I personally don't like this fact, because it introduces ways for the user to break the internal state of the driver. I kind of consider this a functional unsoundness. I found the existing driver very hard to use because it requires me to understand the peripheral, which I as a user don't want to have to understand. I felt I would have to write a driver for the driver to actually use it in a project. 

I would personally rather go the other route - closing it off completely and then step by step adding features as required. But that's just my personal opinion.

Finomnis:
> Here's a rough outline of what it might look like:

Let me think about this. You might be onto something.

Be aware that:
- not every u8 supports dma. It only supports full dma of both the read and write buffer are of the same u32 alignment. Otherwise it will always be the case that we have to decide whether to dma read or write, both aren't possible at they are misaligned.
- there will always be a benefit from using interrupts, even in the dma case. For error handling and for the flush call.
- the few unaligned bytes before and after a transfer also benefit from interrupts, and would still be required on the dma version.

teburd:
I think it’s cool that someone is adding support for eh 1.0. Id note that interrupts are not free of cost and sometimes it really is faster to poll. I saw this in a few spi peripherals in Zephyr that I’ve modified/reviewed. At times interrupts can cause slowdowns. Especially true when doing XIP or on cores with longer pipelines with icache/dcache involved like the M7

Finomnis:
That might be true!

If we
- drop dma support for u8/u16
- assume busy waiting for hal-async's `flush` call is acceptable

Then we can go with @mciantyre's architecture proposal.

Finomnis:
The biggest open question I still have is error handling. But I have to clarify:
- errors can only occur if NOSTALL is activated. Otherwise it's guaranteed that the overflow errors can never occur. So the entire error discussion is irrelevant if we don't allow this flag to be set. I would like to keep the possibility, though. 
- without interrupts, errors would have to be polled repeatedly. According to the RM, if an error occurs, the transmission would have to be aborted immediately. If we relax this requirement for the dma case, this might be possible.

mciantyre:
> I found the existing driver very hard to use because it requires me to understand the peripheral, which I as a user don't want to have to understand.

We support embedded-hal for hiding the peripheral's complexity. We still need to give the user APIs for driver configurations, since embedded-hal doesn't help us here. But after you configure your driver and pass ownership into your embedded-hal-using component, it should become difficult to break the driver's internal state.

I'm sure there's use-cases I'm not considering, but that's the thinking for giving users the lower-level control.

> there will always be a benefit from using interrupts, even in the dma case. For error handling and for the flush call.

We could require the user to transform `InterruptTransmit` into `DmaTransmit`, for example. Once `DmaTransmit` takes ownership of `InterruptTransmit`, it could select the I/O behaviors, wait for flush, check errors.

> I’d note that interrupts are not free of cost and sometimes it really is faster to poll.

We're free to spin in async code if we find it profitable. We block the executor and other tasks.

src/common/lpspi/bus/eh1_impl.rs
```diff
@@ -0,0 +1,81 @@
...
+use super::{Lpspi, LpspiError};
+
+impl<const N: u8> eh1::spi::ErrorType for Lpspi<'_, N> {
+    type Error = LpspiError;
+}
+
+impl<const N: u8, T> eh1::spi::SpiBus<T> for Lpspi<'_, N>
+where
+    T: crate::lpspi::LpspiWord,
+{
```
mciantyre, line 16: All of these implementations use `cassette` to spin on the future until it resolves. `cassette` doesn't care about wakers; it's just spinning, calling `poll()`, which calls a sub-future's `poll()`, which calls a sub-future's `poll()`, which calls... 

Those `polls()` conclude with a call to `StatusWatcherFuture`. This custom future takes a critical section every time we `poll()` it.

My concern is that by spinning on the future, we're frequently entering and exiting a critical section while we're driving the state machine. Could this happen?

examples/rtic_spi.rs
```diff
@@ -4,64 +4,108 @@
...
+                    Operation::DelayUs(10),
+                    Operation::Write(&[420, 69, 42]),
+                    Operation::Write(&[0xFFFF]),
+                    Operation::DelayUs(50),
+                ])
+                .unwrap();
+
+            // To demonstrate larger, DMA based transfers
+            let mut buf = [0xf5u32; 512];
+            spi_device.transfer_in_place(&mut buf).unwrap();
```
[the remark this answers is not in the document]
Finomnis: Oh, sorry, ignore the example. I switched to an external example I didn't upload yet.

No, the error does get caught. The state of the register gets polled in every call, including the poll call of the future. They're is a function that also enters the critical section that does that.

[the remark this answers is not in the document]
Finomnis: If you enable interrupts but don't poll them, then the sync executor will still work but the async one will hang.

src/common/lpspi/bus/eh1_impl.rs
Finomnis: Yes, that does indeed happen. But that shouldn't be a problem because the critical sections are very short.

Finomnis: Is that a problem?

Finomnis: > I also don't see how taking these critical sections during blocking I/O adds value to the user's system

It doesn't. Yes, it would be better to for several reasons to provide a dedicated blocking API. Polling the async function was purely to reduce implementation complexity.

Finomnis: Actually, similar to the waker problem, if we know we can only ever get called by a single context, we could get away with a `TryLock` and an `Atomic<Option<Waker>>`. I might give that a try.

Finomnis: I kind of expected the whole critical section thing to compile to two asm instructions, one disable interrupts instruction and one enable interrupts instruction. Where does the complexity come from?

Finomnis: Correction, I just realized that interrupts also operate in this. So it's not quite as straight forward.

Finomnis:
> But after you configure your driver and pass ownership into your embedded-hal-using component, it should become difficult to break the driver's internal state.

I agree. So we should maybe look into soothing like a builder pattern?

Finomnis:
@mciantyre Might have to take a step back over the next couple of days/weeks. If someone has an idea, be free to use the ideas of my code. There are a couple of things in there that I took a while to figure out, like the proper settings of the timing registers.

src/common/lpspi/bus.rs
```diff
@@ -0,0 +1,344 @@
...
+            res = data.lpspi.watch_for_errors().fuse() => res,
+            res = this.transfer_unchecked(sequence).fuse() => res,
+        };
+
+        cleanup_on_error.finish(result)
+    }
+
+    /// Returns whether or not the busy flag is set.
+    fn busy(&self) -> bool {
+        ral::read_reg!(ral::lpspi, self.lpspi(), SR, MBF == MBF_1)
```
Finomnis, line 282: Turns out this is buggy on imxrt1062 (or all, maybe)? Writing to the FIFO does **not** set `MBF` immediately; if compiled with optimizations this has a chance to report `not busy` although a transfer is in progress.

Finomnis:
@mciantyre Should I close this branch as obsolete?

Finomnis:
Replaced by #209.

PR #146, opened 2023-11-26
Export, demonstrate ENET peripheral

mciantyre:
The `rtic_enet_dhcp_tcp.rs` example runs a DHCP client on an i.MX RT 1170EVK. Once assigned an IP address, it creates a TCP socket that loops back any data sent to it. The example demonstrates some support for [`smoltcp`](https://docs.rs/smoltcp/0.10.0/smoltcp/). It depends on a prototype driver, [imxrt-enet](https://github.com/mciantyre/imxrt-enet), which I'm still thinking through.

The example works on my 1170EVK with my router, but there's still more to do. In particular, I can't bring the speed up to 100Mbit/s; if I try, the receive path doesn't work reliably. MIB statistics show CRC and alignment errors for receive, but transmit is fine. I'm notably **not** bringing up the Ethernet PLL, so next step might be to rule out clock quality issues. Until then, you're playing with a 10Mbit/s driver, enforced by a restricted PHY autonegotiation.


mciantyre:
> I'm notably **not** bringing up the Ethernet PLL, so next step might be to rule out clock quality issues.

I linked with SDK code to simplify PLL1 bringup, remuxed ENET clocks, forced a 100Mbit/s PHY autonegotation, and disabled the 10Mbit/s driver config. There's no more CRC or alignment errors in the receive path, and the example works as expected.

Looks like it's time for the inevitable: make the 1170 clocks useful. What's this analog IP interface thing...?

mciantyre:
A single setpoint also works for PLL1 bring-up, with no analog IP interface needed. The MAC and PHY are up to 100Mbit/s.

mciantyre:
I pushed some changes to the imxrt-enet repo -- a bump in smoltcp version, among others -- that I think break this build. Sorry about that. I'll rebase and grab that later.

PR #148, opened 2023-12-21
GPIO: embedded-hal 1.0 compatibility

Finomnis:


Cargo.toml
```diff
@@ -33,11 +36,19 @@ version = "1"
 package = "embedded-hal"
 version = "0.2"
 
+[dependencies.eh1]
+package = "embedded-hal"
+version = "1.0.0-rc.3"
+optional = true
```
mciantyre, line 42: It's OK with me to make this a non-optional dependency. What's the benefit of keeping it optional?

No matter its mandatory / optional status, we should include some `lib.rs` top-level docs about its support. And if it's optional, users might want to know the `eh1` feature name so they can turn it on.

src/common/gpio.rs
```diff
@@ -37,6 +37,8 @@
 
 use crate::{iomuxc, ral};
 
+use cortex_m::interrupt;
```
mciantyre, line 40: If keeping, prefer the [`critical-section`](https://docs.rs/critical-section/latest/critical_section/) critical section. These drivers can be used on multi-core MCUs, and using `critical-section` better prepares us to support those MCUs.

mciantyre (commented):
Could you share a minimum example that's fixed by 64c3af99f18dc74950dadf8dbfd9c6a30dc12475? Or, could you help me understand why we need this fix?

64c3af99f18dc74950dadf8dbfd9c6a30dc12475 adds critical sections inside routines that (indirectly) require a `&mut Port<...>` exclusive reference. By requiring the exclusive reference, we've made it the user's responsibility to avoid this race condition. If they're sharing their `Port` object across execution contexts, they'll need to wrap it in something like [`critical_section::Mutex`](https://docs.rs/critical-section/latest/critical_section/struct.Mutex.html).

Cargo.toml
Finomnis: If it isn't optional, an `rc.4` might wasm or create without even a code change by us. We have experienced that in RTIC before; I would remove the feature flag once eh1 is properly released. I also wouldn't document it until then. (will probably happen within a couple of weeks)

Finomnis:
> Could you share a minimum example that's fixed by [64c3af9](https://github.com/imxrt-rs/imxrt-hal/commit/64c3af99f18dc74950dadf8dbfd9c6a30dc12475)? Or, could you help me understand why we need this fix?
> 
> [64c3af9](https://github.com/imxrt-rs/imxrt-hal/commit/64c3af99f18dc74950dadf8dbfd9c6a30dc12475) adds critical sections inside routines that (indirectly) require a `&mut Port<...>` exclusive reference. By requiring the exclusive reference, we've made it the user's responsibility to avoid this race condition. If they're sharing their `Port` object across execution contexts, they'll need to wrap it in something like [`critical_section::Mutex`](https://docs.rs/critical-section/latest/critical_section/struct.Mutex.html).

You might be right, this might have been an oversight by my part. I might have carried over some thoughts from LPSPI, where it will be necessary that an interrupt disables itself upon getting fired, making this critical section necessary there.

I will revert that change.

src/common/gpio.rs
Finomnis: Got it, will do so in the future. Will revert that change for now anyway as I realize it isn't necessary (for now, might become necessary when we implement `eh1-async`)

Finomnis:
@mciantyre Ready for more reviews. Any leftover criticism/comments?

Otherwise I'd like to merge this as a base for the other eh1 implementations. Especially as it's only a non-breaking minor change.

Finomnis:
@mciantyre Ping :)

Finomnis:
@mciantyre Ready for review again. Now update to eh-1.0 release.

mciantyre (approved):
I rebased onto commits that fix clippy warnings in other code, simplified the history, and added a CHANGELOG entry.

I dropped the MSRV recommendation. It's not clear why we should target 1.60, or why it's necessary to adopt an MSRV for this PR. And without a CI job, I'm afraid it would be difficult to maintain the MSRV ourselves. If you have a need for an MSRV, we can discuss in a separate issue.

Finomnis:
> I rebased onto commits that fix clippy warnings in other code, simplified the history, and added a CHANGELOG entry.
> 
> I dropped the MSRV recommendation. It's not clear why we should target 1.60, or why it's necessary to adopt an MSRV for this PR. And without a CI job, I'm afraid it would be difficult to maintain the MSRV ourselves. If you have a need for an MSRV, we can discuss in a separate issue.

A agree with all of your points.

PR #149, opened 2023-12-22
Prepare LPSPI driver for `embedded-hal 1.0` rework

Finomnis:
Some fixes/modifications to the existing LPSPI driver.

This is in preparation for the `embedded-hal 1.0` rework, which will introduce a new high level driver that wraps this low level one.

Changes:
- Remove `CS_PIN` from `Spi` device
  - **Rationale:** eh1 primarily uses software CS pins. There is an extension for hardware CS pin support, but I think we should add it later as a special case instead of by default. This will simplify API design a lot.
- Split `Spi` into `Spi` + `SpiCsPin` in the `board` subcrate
- Add initial state to `gpio::Output` constructor. This is important for cases where the state needs to be set **before** enabling the pin.
- Fix/rework check for valid frame sizes (existing implementation did not cover all corner cases)
- Rework SPI clock initialization
  - Make sure we never get a baud rate **faster** than specified (slower is pretty much always ok for peripherals)
  - Configure `PCSSCK` and `SCKPCS` properly for contiguous transfers
- Some improvements in reset logic
- Add `soft_reset` that cancels the current transfer without requiring a full re-initialization
- Move more values into `enqueue_transaction` so that multiple transactions with different settings can be enqueued in series
- Add `flush` function that waits for the current transaction to be finished whilst repeatedly checking for errors
- Fix `disabled()` not waiting for the current transfer to be finished
- Take `set_watermark` out of `disabled`, because it doesn't require it.

### DIFF
--- a/board/src/imxrt1010evk.rs
+++ b/board/src/imxrt1010evk.rs
@@ -168,7 +168,7 @@ impl Specifics {
         let gpio2 = unsafe { ral::gpio::GPIO2::instance() };
         let mut gpio2 = hal::gpio::Port::new(gpio2);
 
-        let led = gpio1.output(iomuxc.gpio.p11);
+        let led = gpio1.output(iomuxc.gpio.p11, false);
         let button = gpio2.input(iomuxc.gpio_sd.p05);
 
         let lpuart1 = unsafe { ral::lpuart::LPUART1::instance() };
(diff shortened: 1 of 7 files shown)


Finomnis:
@mciantyre Sorry that I was gone for a while. Don't know how much time I will be able to spend on the SPI device, but I'll try to do a little :)

Here now with all the compilation issues fixed. Didn't update examples yet, but I think there's no point really in updating them until we have a properly reworked driver.

Any feedback?

Finomnis:
@mciantyre Ping :)

teburd:
I’d say squash this and I’ll merge, changes look good. Ian is presumably busy with other endeavors at the moment.

Finomnis:
@teburd I don't think I agree time to continue this any time soon, so let's keep that one open because it's a breaking change.

mciantyre:
There's useful non-breaking changes in this branch, changes that could help today's users. I extracted them into #156.

Finomnis:
@mciantyre Welcome back! Glad to see your life has cooled down enough to dabble in some Rust again :)

PR #151, opened 2024-01-31
Add support for i.MX RT1050

hatimbt:
Adds basic support to the i.MX RT1050 family of chips.

I have not yet developed hardware tests and examples under `board`, however, I am currently using this HAL on a custom hardware.

The `imxrt-iomuxc` dependency needs the changes in https://github.com/imxrt-rs/imxrt-iomuxc/pull/43 to work.


mciantyre:
Could you give a pass through [the workflow definitions](https://github.com/imxrt-rs/imxrt-hal/blob/main/.github/workflows/rust.yml) and add builds for the new `imxrt1050` feature flag? I think the `lint-log` and `tests` jobs are the only two that need additions.

mciantyre (approved):
Happy to merge after we talk about and merge https://github.com/imxrt-rs/imxrt-iomuxc/pull/43 .

PR #154, opened 2024-03-04
LPUART: embedded-io 0.6 compatibility

Finomnis:


src/common/lpuart.rs
```diff
@@ -930,6 +930,88 @@ impl<P, const N: u8> eh02::blocking::serial::Write<u8> for Lpuart<P, N> {
...
+        // Block until we can write.
+        // In combination with `WriteReady` this call can be non-blocking.
+        while !self.write_ready()? {}
+
+        let mut num_written = 0;
+        for word in buf {
+            self.write_byte(*word);
+            num_written += 1;
+
+            if !self.write_ready()? {
```
teburd, line 962: I take it there's no contractual requirement to write all the bytes for this call?

Finomnis: No, not at all. The contract says "blocking until at least one is written"; same for 'read'.

Finomnis: @teburd The trait provides a [`write_all`](https://docs.rs/embedded-io/latest/src/embedded_io/lib.rs.html#394-403) function that wraps around it if you want the contractual guarantee.

Finomnis:
@mciantyre I have big interest in merging this and then releasing a new version. I have a couple of projects that now point to this branch, because of eh-1's UART and GPIO.

I know SPI isn't there yet, but would it be fine releasing partial eh-1 support already?

teburd (approved):

Finomnis:
I don't have the power to merge this. Is this repository still maintained? @mciantyre doesn't seem to be involved much any more ...

Finomnis:
@teburd went aber you could create a release with the changes?

PR #156, opened 2024-04-30
Add non-breaking LPSPI rework

mciantyre:
These are most of the non-breaking changes proposed in #149. I extracted them so that we could test and release them with today's driver. Unless there's concern, I plan to publish these within the 0.5 releases.


teburd:
looked good to me, glad to see it merged

PR #163, opened 2024-06-08
Backwards compatible eh1 SPI support

Florob:
This adds impls of eh1's `SpiBus` and `SpiDevice` traits to `Lpspi` in a backwards compatible manner.
At least I think it is backwards compatible. There are some added and changes trait bounds, but they ought to be equivalent to or slightly more permissive than the preexisting ones.
The `DelayNs` operation is not supported and documented as such, since without breaking changes we have no delay provider available to us. Users that needs it can use `embedded-hal-bus` in conjunction with the `SpiBus` impl.

This is probably not where we want to go long term. The `Lpspi` type needs to change to support multiple chip selects, which will probably create a more natural split between the Bus and the Device trait. However, I think it is beneficial to provide a basic impl to users in the meantime.

Caveat: I've not yet extensively tested this.


mciantyre (commented):
The backwards compatibility LGTM. Thanks for sharing this idea!

This looks like a good starting point for an EH1 SPI bus and device implementation. I agree that we can break / extend it later.

I pushed my first test case to this branch. All existing transactions in the loopback test are still working on my 1176 MCU. However, the `SpiDevice` transaction sequence

```rust
spi.transaction(&mut [
    TransferInPlace(&mut xfer),
    Read(&mut read),
]).unwrap();
```

isn't finishing the `TransferInPlace` operation. My logic analyzer output is below. It looks like SCK is stuck high, and it's not clocking out the final bit of the final byte.

<details><summary>SCK stuck high before end of transfer</summary>
Screenshot 2024-06-10 at 12 31 31
</details>

Nothing about the continuous transactions looked obviously questionable, so I'm still digging into the issue.

PR #167, opened 2024-06-23
Fix embedded-hal version

Finomnis:
- Update `embedded-hal` to `0.2.6`:
  This crate uses `eh02::blocking::i2c::Operation`, which only exists for `embedded-hal 0.2.6` and upwards
- Update `bitflags` to `1.3`:
  This crate uses `.complement()` which only exists in `bitflags 1.3` and upwards
- Update `ral-registers` to `0.1.2`:
  This crate uses newer features from `ral-registers 0.1.2`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ version = "0.5.5"
 version = "1.0"
 
 [dependencies.bitflags]
-version = "1.2"
+version = "1.3"
 
 [dependencies.fugit]
 version = "0.3"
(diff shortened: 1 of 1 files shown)


mciantyre (approved):
Thanks, looks good. There's [a commit on `main`](https://github.com/imxrt-rs/imxrt-hal/commit/dd0df8dc939232bb8e35a2a9c3cb3254770ac0ad) that uses the trailing comma support in the latest ral-registers release. I bumped the version during my rebase.

PR #170, opened 2024-08-04
Added PLL7 (USB2 PLL) Support

Trevin-Price:
PLL7 is required to use the secondary USB controller. PLL3 is only used for the primary USB controller.

I briefly read through the i.MX RT1060 specification and noticed no meaningful differences between the USB1 and USB2 PLLs. Assuming I haven't overseen anything, PLL7 should have an identical implementation as PLL3, with the only variation being the USB controller number in the register names.


mciantyre:
PLL7 doesn't exist on the smaller i.MX RT MCUs, like the 1010 and 1020. That's why CI is failing.

Consider exporting PLL7 only when building for the 1060 MCUs. It might look similar to how [we include and export PLL1](https://github.com/imxrt-rs/imxrt-hal/blob/06e6f735893eeb6a86f87dee6b9ab06d4e68a2c3/src/chip/imxrt10xx/imxrt1060.rs#L24-L26) when building for the 1060.

Trevin-Price:
My mistake. Thanks for the pointers. Hopefully it's a bit better now.

mciantyre (approved):
All good; thanks for adding this. LGTM, although I'm not testing on hardware. I'll release this in the 0.5 series.

PR #171, opened 2024-08-14
Add Support for FlexCAN(1/2) 

sigil-03:
This PR adds support for the FlexCAN1 and FlexCAN2 interfaces to `imxrt-hal` 0.5.

FlexCAN1 interface has been tested in HW using a Teensy4.1 - I'll have a corresponding PR in `teensy4-bsp` with an `RTIC` example which was used to test. 

More than happy to take any code comments / improvement suggestions! 

Acknowledgements:
The vast majority of this work is from @dstric-aqueduct - who wrote the initial FlexCAN peripheral code in #122.

@sebkuzminsky rebased @dstric-aqueduct's work onto `imxrt-hal` 0.5 which gave me a great starting point to work from.

### DIFF
--- a/src/chip/imxrt10xx/drivers/can/frame.rs
+++ b/src/chip/imxrt10xx/drivers/can/frame.rs
@@ -38,6 +38,20 @@ impl Frame {
         }
     }
 
+    /// Creates a new data frame.
+    #[inline(always)]
+    pub fn new_from_raw_slice(code: u32, id: u32, data: &[u8]) -> Option<Self> {
+        if let Some(data) = Data::new(data) {
+            Some(Self {
+                code: CodeReg::new(code),
+                id: IdReg::new(id),
+                data,
+            })
+        } else {
+            None
+        }
+    }
+
     /// Creates a new remote frame with configurable data length code (DLC).
     ///
     /// # Panics
@@ -232,7 +246,7 @@ impl CodeReg {
     const RTR_MASK: u32 = 0b1_u32 << Self::RTR_SHIFT;
 
     const DLC_SHIFT: u32 = 16;
-    const DLC_MASK: u32 = 0b111_u32 << Self::DLC_SHIFT;
+    const DLC_MASK: u32 = 0b1111_u32 << Self::DLC_SHIFT;
 
     const TIMESTAMP_SHIFT: u32 = 0;
     const TIMESTAMP_MASK: u32 = 0xFFFF_u32 << Self::TIMESTAMP_SHIFT;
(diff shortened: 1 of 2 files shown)


mciantyre (commented):
> More than happy to take any code comments / improvement suggestions!

Our CI workflows will provide you with the fastest feedback. I recommend enabling our GitHub actions on your fork. That way, you can run our automated checks yourself whenever you push to your fork.

One notable issue: builds for the 1011 and 1015 MCUs are [failing](https://github.com/imxrt-rs/imxrt-hal/actions/runs/10398572252/job/28808785504?pr=171#step:4:111). These MCUs do not have a CAN peripheral, so the RAL doesn't export any CAN registers. Since CAN isn't common to all MCUs, we won't be able to place the CAN driver under the `common` module tree. If we export the CAN driver through a chip configuration module (like [imxrt1020](https://github.com/imxrt-rs/imxrt-hal/blob/main/src/chip/imxrt10xx/imxrt1020.rs), [imxrt1060](https://github.com/imxrt-rs/imxrt-hal/blob/main/src/chip/imxrt10xx/imxrt1060.rs), and [imxrt1170](https://github.com/imxrt-rs/imxrt-hal/blob/main/src/chip/imxrt11xx/imxrt1170.rs)), then those build errors should go away.

If you have any questions as you work through CI issues, leave them here and I'll try to help. I'll provide a detailed review once CI is passing. Until then, here's some high-level questions based on the prior PR:

- Does this driver solve the activating interrupt problem noted [here](https://github.com/imxrt-rs/imxrt-hal/pull/122#discussion_r1060106501)?
- How does this driver account for MCU [errata](https://github.com/imxrt-rs/imxrt-hal/pull/122#discussion_r1063067031)?
- Observers seem to want [CAN FD support](https://github.com/imxrt-rs/imxrt-hal/pull/122#issuecomment-1419999202). Is there a path forward for adding that capability?

sigil-03:
Thanks for your comments - I'll start by addressing the failing pipelines, and then I'll have a look at answering your questions about the interrupts, lMCU errata, and CANFD support.

sigil-03:
I've gone through and moved the CAN implementation out of `common` and created a `Drivers` sub-module inside of the `Config` module, which exposes available drivers (in this case just CAN. 

I'm not sure this is the right approach, but it's working. Definitely open to feedback here!

I've also updated the docs + docs example code - so all of the pipeline tests should pass. 

Currently I only expose the CAN interface on the 1060 since the only device I have for testing is a Teensy 4.1

mciantyre (commented):
https://github.com/sigil-03/imxrt-hal/pull/2 has additional thoughts.

SebKuzminsky:
:partying_face:

PR #172, opened 2024-09-29
Lpuart Baud Compute Bug

tprice02:
From my understanding of the common::lpaurt::Baud::compute function, it is meant to test every oversampling rate from 8 to 32, and every modulo divisor from 1 to 8191.

The current code does not reset the modulo divisor value after comparing to an oversampling rate, leading to only one oversampling rate being tested. I've added a test that fails on the current code, but passes in the fixed code, highlighting the issue.

On a separate note, the oversampling rate is tested from 8 to 32, yet the minimum possible value is 4. I suspect that the author of the code found that sampling on both edges of the clock was unstable, or unfavorable for some reason, as this applies at oversampling rates <8. However, starting at the value of 8 limits possible baud rates to a maximum of 3MBd, while I've found higher values to be stable in communication between two Teensy 4.0s (albeit not tested with an oscilloscope). I've thus added a rudimentary fix, where the oversampling rate checking starts at a value of 4 instead of 8 when the baud rate is higher than 3MBd, however it seems simpler to always have it start at 4, so consider this and please state any concerns prior to merging.

### DIFF
--- a/src/common/lpuart.rs
+++ b/src/common/lpuart.rs
@@ -593,8 +593,8 @@ impl Baud {
         let mut best_sbr = 0;
 
         let mut osr = 8;
-        let mut sbr = 1;
         while osr <= 32 {
+            let mut sbr = 1;
             while sbr < 8192 {
                 let b = source_clock_hz / (sbr * osr);
                 let e = max(baud, b) - min(baud, b);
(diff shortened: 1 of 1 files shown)


mciantyre:
Released in 0.5.8

PR #175, opened 2024-11-04
Remove static muts from logging package

mciantyre:
Upcoming clippy versions have more aggressive warnings about static mut, so I'm removing them from the logging package. The defmt frontend uses a custom static with interior mutability. The implementation is otherwise the same.

I refactored the rest of the package to use `&'static mut`s produced by the static_cell package. This lets us remove the explicit uninitialized memory and the mutable statics. The details are unchanged.

Tested the rtic_logging example on an 1170EVK: logging frontend with both backends, and the defmt frontend with just the LPUART frontend.

### DIFF
--- a/logging/Cargo.toml
+++ b/logging/Cargo.toml
@@ -12,6 +12,7 @@ readme = "README.md"
 [dependencies]
 critical-section = "1"
 bbqueue = "0.5"
+static_cell = "2.1.0"
 
 [dependencies.defmt]
 optional = true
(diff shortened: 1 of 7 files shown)


teburd (commented):
This looks quite clean for the most part, and removing the statics seems to have made things nicer.

The enum and match and Inner bit in Poller feels like it *should* be a generic parameter off hand, and not something I really looked hard at before. But I guess if you want to, at run time, switch from uart to usb cdc it makes sense, though most of the time I'd think this is a compile time decision to make?

mciantyre:
It could be a compile-time decision, and most examples in this repo demonstrate backend selection based on a constant. But by representing a `Poller` with runtime dispatch, we enable a use-case where programs can select a logging backend during start-up (read a config from storage, check a GPIO, etc). These users don't need to propagate a backend type state around their interfaces.

PR #178, opened 2024-11-22
Reduce common drivers to support 1180 MCUs

mciantyre:
The 1180 has different GPIO and PIT peripherals. There's no SVNS equivalent (or there is and I haven't found it). Lower those modules and their dependent modules into `chip`, and re-export them for all supporting MCUs.

Once we reduce what's common, we can start to support the 1180. I added a CI job to show a successful build. See imxrt-rs/imxrt-ral#52 for the accompanying imxrt-ral changes.


teburd (approved):

teburd:
Makes sense, I guess questions start coming up with how does the mcux sdk deal with these peripheral block revisions/differences, I guess some ifdef's in C...

korken89:
Thank you for starting too look into this @mciantyre !
We've been a little swamped, but will have a look ASAP.

Issue #142, opened 2023-11-05
Road to `embedded-hal 1.0`


Finomnis:
This is a tracking issue of our progress and tasks to integrate the upcoming `1.0` version of `embedded-hal`.

TODO: Link the migration guide here

Tasks:
- [ ] Add a task here.

mciantyre:
No preference for or against breaking changes. Folks should feel free to explore new driver designs if they're warranted for embedded-hal 1.0. I'd still like to maintain some kind of embedded-hal 0.2 support. We could implement both traits with our drivers, or we recommend something like [embedded-hal-compat](https://docs.rs/embedded-hal-compat/latest/embedded_hal_compat/).

It might also be nice to support the async embedded-hal traits. My past attempts with async embedded i.MX RT drivers taught me what _not_ to do, so I'm looking for help here. I hope to see approaches that are decoupled from any async runtime, though I'm not sure if that's possible or a good idea.

Finomnis:
My plan is currently to implement purely the `async` version and convert them to blocking with the `cassette` crate. I do have quite some experience with async (I'm the owner of `tokio-graceful-shutdown` and rewrote tokio's `CancellationToken`), and I'm motivated to answer questions if someone needs help. :)

Finomnis:
I do agree with the plan to decouple from a specific runtime, and I don't think it's hard. So far for my lpspi rework I don't have anything that is specific to rtic or similar. E-h 1.0 so far seems will thought through.

Finomnis:
@mciantyre For timer adapters, can't we use [`rtic-monotonics`](https://docs.rs/rtic-monotonics/latest/rtic_monotonics/imxrt/index.html)? While having `rtic` in its name, it's actually runtime agnostic.

Finomnis:
@mciantyre Would you mind an intermediate release, now that implemented GPIO and UART for e-h 1.0? I need that for a project of mine and would like to official release it, but it depends on the github version of `imxrt-hal` because of this.

prutschman:
Apologies if this is the wrong forum. I see that the roadmap checkbox indicates timer work is done, and that adapters are deferred to rtic-monotonic. Blocking appears to impl eh02::blocking::delay::{DelayMs,DelayUs}, but not eh1::delay::DelayNs. Is this an oversight, or am I missing something about the new way of doing things? I didn't see anything about DelayNs in rtic-monotonic.

Finomnis:
[...]

PR #181, opened 2024-12-01
Add SpiBus, SpiDevice implementations

mciantyre:
We implement all I/O in terms of `transact`. We'll manually flush at the
end of eh02 SPI writes. Since the `SpiBus` interface exposes a flush, we
have no need to perform an internal flush when transacting I/O.

Keep the spinning async state machines to manage the TX and FX FIFOs.
We introduce a schedule to eagerly execute transmit operations ahead of
receive operations. Without this ability, we'll stall the LPSPI bus.
(The NOSTALL field can change this behavior, but then we move complexity
into TX FIFO underrun handling.) There's some miri tests to simulate our
hardware utilization.

Our `SpiDevice` impls mimic the embedded-bus design. There's an
"exclusive" device that owns the bus and uses a hardware chip select.
There's a `RefCellDevice` that lets users share the bus and use hardware
chip selects. There's no critical section / mutex device, but it should
be straightforward to add that later.

Part of #142.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -118,6 +118,24 @@ jobs:
     - name: Run unit, integration tests
       run: cargo test --features=${{ matrix.chips }} --tests --package=imxrt-hal --package=imxrt-log
 
+  # Make sure our unit tests can pass (some recent version of) miri.
+  miri:
+    needs: tests
+    strategy:
+      matrix:
+        chips:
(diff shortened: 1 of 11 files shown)


Finomnis:
Thanks for working on this, I wasn't able to continue my work due to personal things, sorry :) glad to see you are still motivated <3

PR #188, opened 2025-06-14
Sai RX implementation and Teensy audio board example

libesz:
Exposed sai config properties as public, so the user can update before split().

rtic_sai_pcm5102 example enables the Rx+Tx clock in sync, so Tx data works with Rx clock (e.g. to match Teensy audio board pinout).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -219,7 +219,11 @@ required-features = ["board/imxrt-log","board/sai"]
 
 [[example]]
 name = "rtic_sai_pcm5102"
-required-features = ["board/imxrt-log","board/sai"]
+required-features = ["board/imxrt-log","board/sai","board/teensy4"]
+
+[[example]]
+name = "rtic_sai_sgtl5000"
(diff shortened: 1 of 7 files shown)


libesz:
@teburd @mciantyre Hey! Please let me know what you think about this PR. The main purpose is to enable the teensy audio board usage with the original pinout (e.g. put straight on top of the teensy board). That requires to enable the Rx side of the SAI (as the audio board is using the Rx clock pins) even if only Tx is used. I came up with this initial idea to make the sai_config properties public, so the user can tweak it for their need. Not sure if we want something more strict, like setters for all properties or higher level methods like "set rx+tx up with synced clocks".

libesz:
@mciantyre let me ping this PR review. :bow:

teburd:
I’ll give @mciantyre a week to review, if no comments come I’ll merge. Can always correct issues as needed

mciantyre (approved):

PR #191, opened 2025-07-18
Add vmu rt1170 board

SpinFast:
Adds the initial vmu rt1170 board and needed FCB configuration for booting.

Needs imxrt-rs/imxrt-boot-gen FCB for this board first with it published to crates.io


mciantyre (commented):
Looks like this board should be able to, at least, turn on its LED. Mind adding [just enough coverage](https://github.com/imxrt-rs/imxrt-hal/blob/1f3d6fb8d36e920586a934a8644e62a7dbab19f4/.github/workflows/rust.yml#L71-L95) for the board and the `hal_led` example? I think this should do it:

```
# We only care that the VMU board can turn on its LED
- --features=board/vmu-rt1170-cm7 --example=hal_led # Other examples, maybe...
```

I'm prioritizing a CI build of just the board. If there's other examples that are working, it could be nice to add them, but also OK if that doesn't happen / happens alter.

---

crates.io should have all dependent crates. Everything else LGTM, but I'm not doing a detailed review to check pinouts / hardware capabilities.

mciantyre (approved):

PR #198, opened 2026-03-15
Remove driver type states

mciantyre:
The `with_pins` driver constructor will still type check pin-to-peripheral compatibility, similar to how `new` did this. But after construction, we discard the instance number and any pin types. Although we [can't release drivers anymore](https://github.com/imxrt-rs/imxrt-hal/issues/20), I found this makes it easier to use most drivers. Folks who need to release & configure pins / peripherals can reconstruct their components -- likely using some `unsafe` calls -- after ensuring there's no aliasing.

`without_pins` remains for folks who weren't using these type-checked APIs. These users are broken by the missing instance numbers. Drivers that still need to check their pin compatibility after construction do this at runtime.


tacertain:
Hopefully this is more helpful than annoying, but here's what Claude thought:

# PR #198: Remove driver type states — Review

## Overview

This is a large, well-structured refactoring (58 files, ~2k additions, ~2.3k deletions, net reduction) that removes `const N: u8` instance generics and pin type parameters from all peripheral drivers. The pattern is consistent: drivers go from `Driver<P, const N: u8>` to simply `Driver`, with `with_pins()` / `without_pins()` constructors that consume pins and discard type info via a new `into_any()` helper.

## What works well

- **Consistent pattern across all drivers** — LPI2C, LPSPI, LPUART, GPT, PIT, ADC, SAI, GPIO, RGPIO, and FlexPWM all follow the same transformation.
- **User-facing simplification is significant** — Type aliases go from `Lpuart<ConsolePins, 2>` to just `Lpuart`. Board support code and examples become substantially more readable.
- **Fixes real bugs in SAI** — The commit message for "Remove SAI type states" documents several pre-existing bugs being fixed:
  - FPACK values for 8-bit (`0b01` was reserved; corrected to `0b10`) and 16-bit (`0b10` corrected to `0b11`).
  - TX path never configured FPACK at all (always wrote `0_u32`); now writes `packing as u32`.
  - `without_pins` never tracked TX/RX pin masks, breaking subsequent `split()` calls.
- **Typo fix**: `set_interrupts(interrutps)` → `set_interrupts(interrupts)` in SAI.
- **New FlexPWM output masking APIs** are a nice addition.

## Items worth discussing

### 1. `into_any()` safety (`src/lib.rs:327`)

```rust
unsafe {
    let block: *const T = &*inst;
    AnyInstance::new(block)
}
```

This reconstructs an `Instance` from a raw pointer, discarding the original lifetime proof. The safety comment is adequate ("user who made inst claims it points to static MMIO"), and since `Instance` creation is already `unsafe` with that same contract, this holds. But a doc comment on the function (not just an inline safety comment) would help future maintainers understand the invariant.

### 2. FlexPWM `set_output_enable` parameter reorder

The signature changed from `(channel, mask)` to `(mask, channel)`. Looking at the call sites, the new API reads as `pwm.set_output_enable(SM2.mask(), A)` which is arguably more natural ("enable this mask on channel A"). The reorder is fine, and since both parameters are distinct types (`Mask` vs `Channel`), this won't silently miscompile.

### 3. PIT `lifetime_value()` preconditions

This method assumes channel 1 is chained to channel 0 and both are enabled, but there's no runtime check or documented panic/safety section. If violated, it silently returns garbage. Consider either:

- Returning `Option<u64>` (check chaining state), or
- At minimum, documenting the precondition under a `# Panics` or `# Correctness` heading.

### 4. SAI loss of compile-time packing/word-size validation

Old code: `Packing<WORD_SIZE>` trait prevented `Packing16bit` with 8-bit words at compile time. New code: `split(8, 2, Packing::Pack16bit, ...)` compiles but is an invalid configuration. This is an inherent tradeoff of the PR's approach. Consider a `debug_assert!` in `split()` to catch mismatches (e.g., `Pack8bit` requires `word_size == 8`).

### 5. GPIO `duplicate_instance()` aliasing

`Port::output()` and `Port::input()` create new `AnyInstance` values pointing to the same register block. The safety comment claims atomic access only. This is correct for GPIO (set/clear/toggle registers are atomic), but worth a brief audit to confirm no read-modify-write patterns exist on shared registers in `Output`/`Input` methods.

### 6. Minor: stale doc comment in LPUART

A comment likely still says "When `new` returns" but the constructor is now `with_pins`. Should say "When this call returns" or similar.

## Verdict

This is a high-quality refactoring with a clear thesis: trade some compile-time guarantees for dramatically simpler types that don't force generics to propagate through user code. The pattern is applied consistently, the SAI commit fixes real hardware bugs, and the examples demonstrate the improved ergonomics well.

**Suggested before merge:**

1. Document `lifetime_value()` preconditions (PIT)
2. Consider `debug_assert!` for packing/word-size mismatch in SAI `split()`
3. Minor doc fixups (LPUART "new" reference, `into_any` doc comment)

None of these are blockers — the PR is in good shape.

mciantyre:
> Testing status on my Teensy4.1.

Thanks for taking on that testing! Appreciate the help. I'll test the rest with my other boards, enough to get coverage on the missing drivers. Should start tomorrow.

---

> Suggested before merge:
> 
> 1. Document lifetime_value() preconditions (PIT)
> 1. Consider debug_assert! for packing/word-size mismatch in SAI split()
> 1. Minor doc fixups (LPUART "new" reference, into_any doc comment)

These are good. I'll grab them in follow-on commits, along with the "if error, return pad ownership" idea.

> 3. PIT lifetime_value() preconditions [...] Returning Option<u64> (check chaining state)

I'd like reading the timestamps it to be minimal cost. I'd support a separate method, like `fn is_lifetime_enabled(&self) -> bool` that users could call themselves, if they really wanted.

src/chip/drivers/pit.rs
```diff
@@ -181,275 +129,159 @@ impl<const CHAN: u8> Pit<CHAN> {
...
-        // Safety: there can only be one (safe) instance of this chained timer.
-        // We effectively own these registers. These addresses are valid MMIO
-        // registers.
-        let mut high = self.low.instance.LTMR64H.read();
-        let mut low = self.low.instance.LTMR64L.read();
+        let mut high = self.pit.LTMR64H.read();
+        let mut low = self.pit.LTMR64L.read();
 
-        let ldval0 = self.low.load_timer_value();
+        let ldval0 = self.load_timer_value(Channel::Chan0);
```
mciantyre: The errata workaround is

> In lifetimer mode if the read value of LTMR64L[LTL] is equal to LDVAL0[TSV] then read both
LTMR64H and LTMR64L registers one additional time to obtain the correct lifetime value.

However, `self.load_timer_value` is implemented as

```rust
    pub fn load_timer_value(&self, channel: Channel) -> u32 {
        ral::read_reg!(ral::pit::timer, self.timer(channel), LDVAL).saturating_add(1)
    }

```

There's an off-by-one error. This method should read `LDVAL` directly, without that increment.

mciantyre: https://github.com/imxrt-rs/imxrt-hal/pull/200 backports a fix to the 0.5 release series.

mciantyre:
Skipped in Teensy 4.1 testing, but covered on my 1010EVK:

- [x] hal_clock_out - Console menu works. Not connecting my tools to probe the `CLKOUT` pads.
- [x] hal_i2c_lcd1602
- [x] hal_ocotp_fuse - Read path worked. Driver wasn't touched by refactor, so not testing further.
- [x] rtic_defmt_rtt
- [x] rtic_sai_wm8960 - Jammed headphones into board, heard a tone. Didn't sound like 440Hz to me, but it's consistent with tone on `main`.
- [x] rtic_usb_serial
- [x] rtic_usb_test_class - Benchmark results below, just for fun. Results consistent with `main`.

<details>
<summary>rtic_usb_test_class results on 1010EVK</summary>

```
❯ echo "In the usb-device repo"
In the usb-device repo

❯ git rev-parse --short HEAD
a6e6c91

❯ cargo test --features=test-class-high-speed --test=test_class_host
   Compiling usb-device v0.3.2
    Finished [`test` profile [unoptimized + debuginfo]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 0.88s
     Running tests/test_class_host/main.rs (target/debug/deps/test_class_host-8e50bbd26accd352)
test_class_host starting
looking for device...

running 11 tests
test control_request ... ok
test control_data ... ok
test control_data_static ... ok
test control_error ... ok
test string_descriptors ... ok
test interface_descriptor ... ok
test iso_endpoint_descriptors ... ok
test bulk_loopback ... ok
test interrupt_loopback ... ok
test bench_bulk_write ... ok
  16 transfers of 65536 bytes in 0.135s -> 62.073Mbit/s
test bench_bulk_read ... ok
  16 transfers of 65536 bytes in 0.077s -> 108.579Mbit/s
0 failed, 11 succeeded

ALL TESTS PASSED!
```

</details>

I think it's fine to keep skipping

- [ ] ~hal_i2c_mpu9250~ - Inventory allocated, but LPI2C covered by SAI & LCD1602 testing.
- [ ] ~rtic_sai_pcm5102~ - Not equipped, but SAI covered elsewhere.

mciantyre:
I'll keep this open through the weekend for folks who want to provide feedback. Otherwise, I intend to merge on Monday.

PR #199, opened 2026-03-17
Fixups from testing

tacertain:
These are a collection of little fixes to examples that I found while testing


mciantyre (approved):
Thanks for working through that button troubleshooting.

